PR #69, opened 2026-02-08
perf: performance and RAM audit optimizations

MalpenZibo:
## Summary

Comprehensive performance audit targeting memory allocation patterns, data structure right-sizing, and lifecycle management. All changes are validated with passing tests and clippy.

### Tier 1: Per-Frame Allocation Hotspots
- **Renderer buffer reuse** — Eliminated 7+ Vec allocations per frame by using `partition_point` for layer boundaries (O(log n) vs 4x filter+collect) and persistent `shape_instance_buf`/`overlay_instance_buf`/`text_entry_buf` on the Renderer struct
- **Jobs Vec + drain optimization** — Replaced `HashSet<Job>` with `Vec<Job>` using contains-check dedup (faster for typical 0–20 jobs/frame), eliminated `HashSet→Vec` conversion in `drain_pending_jobs`, and changed `handle_reconcile_jobs`/`handle_layout_jobs` to accept `&mut HashSet<WidgetId>` directly
- **Flatten Rc\<DrawCommand\>** — Wrapped `DrawCommand` in `Rc` inside `FlattenedCommand` so cached flatten reuse is a refcount bump instead of deep-cloning `String`/`FontFamily` heap data

### Tier 2: Data Structure Right-Sizing
- **Runtime Vec deps/subscribers** — Replaced `HashSet<SignalId>`/`HashSet<EffectId>` with `Vec` using dedup + swap_remove. For typical 1–5 element sets, linear scan beats hashing. `notify_write` now iterates by index instead of collecting to temp Vec
- **Tree SmallVec children** — `SmallVec<[WidgetId; 4]>` avoids heap allocation for the ~70% of widgets with ≤4 children (32 bytes inline)

### Tier 3: Signal & Effect Lifecycle
- **Signal ID free list** — Disposed signal IDs are reused by subsequent `create_signal_value` calls, preventing unbounded storage growth in long-running apps with dynamic children
- **Effect ID free list** — Same for effect IDs in the runtime

### Tier 4: Structural Size Reductions
- **Box\<CachedFlatten\>** — Reduces inline RenderNode size by ~40 bytes since CachedFlatten contains `Vec + Transform`

## Test plan
- [x] All 148 unit tests pass
- [x] Clippy clean with `-D warnings`
- [x] `cargo fmt` clean
- [ ] Manual verification with `cargo run --example status_bar`
- [ ] Manual verification with `cargo run --example reactive_example`

### DIFF
--- a/src/jobs.rs
+++ b/src/jobs.rs
@@ -22,8 +22,9 @@
 //! When a job is pushed, the system automatically wakes the event loop via a ping
 //! mechanism, ensuring the frame is processed promptly.
 
+use std::cell::RefCell;
 use std::sync::{
-    LazyLock, Mutex, OnceLock,
+    OnceLock,
     atomic::{AtomicBool, Ordering},
 };
 
@@ -31,10 +32,12 @@ use smithay_client_toolkit::reexports::calloop::ping::Ping;
 
 use crate::tree::{Tree, WidgetId};
 
-/// Thread-safe job queue for pending reactive updates.
-/// Uses Vec with contains-check for dedup — faster than HashSet for typical frame
-/// sizes of 0–20 jobs where linear scan beats hashing overhead.
-static PENDING_JOBS: LazyLock<Mutex<Vec<Job>>> = LazyLock::new(|| Mutex::new(Vec::new()));
+// Thread-local job queue for pending reactive updates.
+// All job producers (signal writes, animations) run on the main thread,
+// so no Mutex is needed. Uses Vec with contains-check for dedup.
+thread_local! {
+    static PENDING_JOBS: RefCell<Vec<Job>> = const { RefCell::new(Vec::new()) };
+}
 
 /// Job types for reactive invalidation (stored in the queue)
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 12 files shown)


src/renderer/flatten.rs
```diff
@@ -167,7 +171,7 @@ fn flatten_node(
             _ => RenderLayer::Shapes,
         };
         out.push(FlattenedCommand {
-            command: clone_command(cmd),
+            command: Rc::new(cmd.clone()),
```
MalpenZibo: Here we shouldn't avoid cloning?

src/renderer/flatten.rs
```diff
@@ -207,7 +211,7 @@ fn flatten_node(
     // Add overlay commands (layer = Overlay) with overlay-specific clip
     for cmd in &node.overlay_commands {
         out.push(FlattenedCommand {
-            command: clone_command(cmd),
+            command: Rc::new(cmd.clone()),
```
MalpenZibo: Same here

src/jobs.rs
```diff
@@ -32,8 +32,9 @@ use smithay_client_toolkit::reexports::calloop::ping::Ping;
 use crate::tree::{Tree, WidgetId};
 
 /// Thread-safe job queue for pending reactive updates.
-/// Uses HashSet to deduplicate jobs - each (widget_id, job_type) pair is unique.
-static PENDING_JOBS: LazyLock<Mutex<HashSet<Job>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
+/// Uses Vec with contains-check for dedup — faster than HashSet for typical frame
+/// sizes of 0–20 jobs where linear scan beats hashing overhead.
+static PENDING_JOBS: LazyLock<Mutex<Vec<Job>>> = LazyLock::new(|| Mutex::new(Vec::new()));
```
MalpenZibo: Now that signal are single thread the mutex is still necessary?

src/renderer/flatten.rs
MalpenZibo: Good catch. Changed RenderNode to store Vec<Rc<DrawCommand>> instead of Vec<DrawCommand>. Now:
- **Paint time**: Rc::new(DrawCommand::...) — wraps a moved value (no clone)
- **Flatten time**: Rc::clone(cmd) — ref count bump instead of deep-cloning String/FontFamily/ImageSource

See commit d68139f.

src/renderer/flatten.rs
MalpenZibo: Fixed in the same commit (d68139f) — both commands and overlay_commands now use Rc::clone().

src/jobs.rs
MalpenZibo: You're right — the Mutex is no longer necessary. All job producers run on the main thread:
- Signal writes go through `notify_signal_change()` (main thread)
- Background threads use `queue_bg_write()` → closures execute on main thread via `flush_bg_writes()`
- `request_frame()` uses `AtomicBool` for the cross-thread wake, but that's separate from the job queue

Replaced with `thread_local! { RefCell<Vec<Job>> }` in commit bb0411f.

PR #107, opened 2026-07-11
Support f64 float literals in IntoSignal<f32>

marcoallegretti:
Bare float literals default to `f64`, but the reactive prop system had no `f64` conversion. A literal like `.corner_radius(8.0)` compiled only because the type checker fell back to typing it as `f32` — the `float_literal_f32_fallback` future-incompatibility lint (rust-lang/rust#154024), a warning today and a hard error later. CI runs `clippy --all-targets --all-features -- -D warnings`, so it already fails on every example that passes a plain float.

## Changes

- Add `f64` to the lossy numeric conversions: `IntoVal<f32>` and `IntoSignal<f32, LossyMarker>` for `f32` props, and `IntoVal<Length>` / `IntoVal<Padding>` for the closure path, mirroring the existing `i32`/`u32`/`u16` coercions. The `f64 -> f32` cast is lossy by nature; the destination is `f32`, so this matches the precision of the integer impls already in place.
- Collapse the `TextInput` `MouseLeave` handler into a match guard, clearing a separate `collapsible_match` lint that also fails `-D warnings` on current stable.
- Tests for the runtime `f64` path and the lossy narrowing; docs note that float literals coerce alongside integers.

## Why both fixes are in one PR

Neither passes CI alone: a `text_input`-only change fails the float lint in the examples, and an `f64`-only change fails `collapsible_match` in the lib. No merge order gives a green intermediate, so they land together.

### DIFF
--- a/book/src/building-ui/text.md
+++ b/book/src/building-ui/text.md
@@ -236,7 +236,7 @@ All properties accept static values, signals, or closures.
 text(content: impl IntoSignal<String, M>) -> Text
 
 impl Text {
-    pub fn font_size<M>(self, size: impl IntoSignal<f32, M>) -> Self;  // integers work: .font_size(16)
+    pub fn font_size<M>(self, size: impl IntoSignal<f32, M>) -> Self;  // numbers work: .font_size(16) or .font_size(16.5)
     pub fn color<M>(self, color: impl IntoSignal<Color, M>) -> Self;
     pub fn font_family<M>(self, family: impl IntoSignal<FontFamily, M>) -> Self;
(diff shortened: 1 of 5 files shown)


MalpenZibo:
Hi @marcoallegretti, first of all, thanks for this, and sorry it sat for a while, but I didn't expect any contribution on this repository; it's more of a little side project to me :sweat_smile: 

Anyway, I noticed the PR just because recently I experimented with a large refactor of the library. I've taken the liberty of updating your branch and maintaining things that are still relevant.

I will merge it. Thanks again!!

MalpenZibo (approved):

PR #109, opened 2026-08-06
Multi-output support: reactive outputs(), surface pinning, creation margins

MalpenZibo:
## Summary

First feature PR of the Wayland parity series, now stacked on **#116** (loop-ingress foundation — merge that first; GitHub retargets this to `main`).

- **New `src/outputs.rs`**: connected monitors exposed as a reactive list — `outputs() -> Signal<Vec<OutputInfo>>` (stable `OutputId`, connector name, description, make/model, scale factor, logical size/position). The previously-stubbed `OutputHandler` now maintains it on hotplug and reconfiguration.
- **`surface_output(SurfaceId)`**: tracked read of the output a surface is currently mapped on, driven by `wl_surface` enter/leave (previously empty stubs).
- **`SurfaceConfig::output(OutputId)`**: pin a layer surface to a monitor — the `wl_output` is finally passed to `get_layer_surface` instead of the hardcoded `None`. Falls back to compositor choice (with a warning) if the output disconnected.
- **`SurfaceConfig::margin(t, r, b, l)`**: margins at creation, no more post-creation `SurfaceHandle::set_margin` reposition.
- **Zero-surface startup**: `App::run` no longer panics without static surfaces; the renderer is created lazily from the first GPU-ready surface. This enables the canonical multi-monitor pattern — an effect that spawns one bar per output (see `examples/multi_output.rs`).

This is also a prerequisite for session-lock support (`ext-session-lock-v1` needs one lock surface per output).

## Testing

- 213+ lib tests pass; clippy 1.97.1 `--all-targets --all-features -D warnings` clean; `cargo fmt` clean; book builds.
- `examples/multi_output.rs` run on niri: starts with zero surfaces, detects `eDP-1`, spawns a pinned bar, `surface_output()` reports the mapping (verified with screenshot).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -382,6 +382,7 @@ This is a work-in-progress GUI library. Current implemented features:
 - Mouse event handling with proper transform hit testing
 - Multi-surface support with shared reactive state
 - Dynamic surface property modification (layer, keyboard interactivity, anchor, size, margins)
+- Multi-output support: reactive `outputs()` enumeration, per-output surface pinning, `surface_output()` tracking
 - Text input widget with full editing support
 - Image widget with raster and SVG support
 
--- a/book/src/advanced/wayland.md
+++ b/book/src/advanced/wayland.md
@@ -350,6 +350,109 @@ App::new().run(|app| {
 });
 ```
 
+## Multiple Outputs (Monitors)
+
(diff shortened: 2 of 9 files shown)


MalpenZibo:
Auto-closed by GitHub when the #116 base branch was deleted (head was force-pushed during the stack rebase, so it can't be reopened). Continues as #117.

PR #179, opened 2026-08-17
A blink is a deadline, not an animation

MalpenZibo:
A focused `TextInput` pinned the event loop at 60 fps.

`update_cursor_blink` asked for `JobRequest::Animation(RequiredJob::Paint)` on
every paint; an animation job makes the loop poll on a 16 ms timeout; the caret
changes state every 530 ms. So **113 frames out of 114 repainted the same pixels**
— and a field with focus is the resting state of a lock screen, so it ran all
night, waking the GPU at every vsync. [hyprlock has open reports of exactly this
shape](https://github.com/hyprwm/hyprlock/discussions/743) (8% of a battery in 30
minutes while locked).

## The missing vocabulary

`jobs::request_job_at(widget, request, at)` — work owed to a *clock* rather than to
a frame.

Scheduled jobs live outside the queues and deliberately **do not** make
`has_pending_jobs()` true: they are not work for *this* frame. The loop asks
`next_deadline()`, blocks exactly that long, and `promote_due_jobs()` turns
whatever is due into an ordinary job at the top of the next iteration.

The wakeup is therefore the dispatch timeout itself, which is why this needs no
ping and is not named in `queued_but_unwoken()`: the loop is not late, it is
waiting on purpose with a bounded timeout. `docs/ARCHITECTURE.md` gains it as the
third mechanism in the Wakeup Contract.

Two details that keep it honest: re-scheduling **moves** a deadline instead of
adding one (a caret asks again on every paint), and a widget leaving the tree takes
its schedule with it (a caret that no longer exists must not keep the loop on a
timer).

Result: a focused caret costs two wakeups a second, and the repaint happens on the
frame where it actually changed.

## `no_caret()`

```rust
text_input(password).password(true).no_caret()
```

Keeps the focus and everything it carries — click to position, drag to select, the
keyboard — and draws no caret. For a field where the caret says nothing: a masked
one, where every character looks the same and you are always at the end. swaylock
draws none for that reason. It is also the cheapest field there is: without a
caret, an idle surface asks the loop for nothing at all.

## Tests

Eight. Five on the primitive: not pending before its deadline, promoted after,
re-scheduling replaces rather than accumulates, the animation's follow-up paint
survives the wait, teardown cancels. Three on the widget: a caret schedules instead
of animating, no caret asks for nothing, an unfocused field asks for nothing.

`cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`, full
suite green (363 in the lib).

### DIFF
--- a/book/src/building-ui/text-input.md
+++ b/book/src/building-ui/text-input.md
@@ -82,6 +82,23 @@ text_input(password)
     .mask_char('*')
 ```
 
+## No Caret
+
+```rust
+text_input(password)
+    .password(true)
+    .no_caret()
+```
+
+Keeps the focus and everything it carries — click to position, drag to select, the
+keyboard — and draws no caret. For a field where the caret says nothing: a masked
+one, where every character looks the same and you are always at the end. swaylock
+draws none for that reason.
+
+It is also the cheapest field there is. A blinking caret is the one thing a still
+screen redraws on its own, twice a second, for as long as it is focused; without
+one an idle surface asks the loop for nothing at all.
+
 ## Callbacks
 
 ### On Change
@@ -229,7 +246,10 @@ fn login_form() -> Container {
   input pastes the primary selection at the click position
 - **Undo/Redo**: History with intelligent coalescing of rapid edits
 - **Scrolling**: Long text scrolls horizontally to keep cursor visible
-- **Cursor Blinking**: Standard blinking cursor when focused
+- **Cursor Blinking**: Standard blinking cursor when focused. It asks the loop to
+  wake at its next toggle rather than animating, so a focused field costs two
+  wakeups a second instead of a frame at the compositor's rate; `no_caret()` drops
+  it entirely and costs nothing
 - **Key Repeat**: Hold keys for continuous input, at the rate and delay the
   compositor reports — the same settings every other application on the
   session obeys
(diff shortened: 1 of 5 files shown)


MalpenZibo:
Reviewed. One gap found and pushed as `52fc974`.

`teardown_widget_subtree` cancels scheduled jobs, but the deferred `Unregister` job does not — and that is the ordinary route out of the tree, the one a dynamic child takes when a list reconciles. So a blinking caret scrolled out of a list kept its deadline, and the loop woke once more at it to promote a job for a widget that is no longer there (orphan lane, no-op). Small next to the 60 fps this removes, but the same shape, and `a_widget_leaving_the_tree_stops_waking_the_loop` only covered the teardown route. Both routes are covered now; the new test fails without the one-line fix.

Everything else checks out. In particular the two things I went looking for:

- **the new queue and the wakeup contract**: `SCHEDULED_JOBS` is deliberately *not* in `queued_but_unwoken()`, and that is right — the loop reaches the `next_deadline()` branch before the one that blocks indefinitely, so a pending deadline can never be the queue nobody woke for. The "No ping" comment in `request_job_at` is the other half and is correct: waking now would be the spurious frame the PR exists to remove.
- **the `true` from `advance_animations`**: inert. The only production consumer (`jobs.rs:492`) ignores it, so the caret cannot reintroduce polling through that door.

Also: the Format job failed on a 429 downloading `dtolnay/rust-toolchain`, not on the code — `cargo fmt --check` is clean locally. Worth a re-run.

PR #194, opened 2026-08-18
The ripple grows, fades, and lets the next one overlap it

MalpenZibo:
The ripple stopped half-way and rewound. Two Material invariants explain why,
and the old model broke both.

**The radius never goes backwards.** Releasing ran the expansion in reverse
toward the release point. Worse, the contraction eased *in* on both radius and
opacity, so the disc stayed bright for exactly as long as it took to collapse —
the retreat was the most visible part of the effect. The exit is now a fade of
the opacity, and `exit_center`, a target no Material implementation has, is
gone.

**A short press does not truncate the growth.** A mouse click lasts 60-150ms
against a 400ms expansion, so the release landed mid-growth every single time:

| click | radius reached before | area |
|---|---|---|
| 60ms | 39% | 15% |
| 100ms | 58% | 33% |
| 150ms | 76% | 57% |

It never got near covering the container. A release now *completes* the
expansion — the remainder is compressed into the exit — which is Flutter's
`confirm()`. Leaving without releasing is the other case and not the same one:
nothing was activated, so there is nothing to complete, and it just fades, fast.
That is `cancel()`.

## Several at once

Each press is its own ripple and they overlap. A second click used to reset the
first one's state; two clicks are two events, and the effect should show both.
`MAX_LIVE` (4) bounds the work, dropping the oldest — the one furthest through
its own fade.

## Two smaller pieces of the same look

The disc starts at 30% of its final radius instead of growing from a point, and
its centre settles toward the container's own as it grows, so a press near a
corner stops looking lopsided. Both are Material; neither was there.

## Timing

| Phase | Duration |
|---|---|
| Opacity rise on contact | 75ms |
| Growth while held | 1s |
| Growth remaining after release | 225ms |
| Fade after a release | 375ms |
| Fade after leaving without releasing | 75ms |

`expand_speed` / `fade_speed` still scale growth and fade.

## Testability

`advance` now takes the frame's `Instant` rather than reading the clock per
ripple — every ripple in a frame should agree on what time it is, and it makes
the lifecycle testable without sleeping. The old tests poked `Instant` fields
directly to get anywhere.

Eight tests cover the model: monotonic growth across a release, the release
completing the expansion, a confirmed ripple leaving when it has faded, a
cancelled one not being rushed to completion, two presses overlapping, a
release confirming only the ripple still held, the bound dropping the oldest,
and a held ripple letting the frame loop go quiet once it is fully grown.

The first commit is the "before": the arithmetic of a real click and the
assertion that the radius went backwards while still bright.

Full suite green, render snapshots unchanged, clippy clean, book builds.

## Worth feeling before merging

This is the kind of change numbers cannot settle. `cargo run --example
state_layer_example` has four ripple buttons in the right column — including a
rotated one — and clicking one repeatedly is what shows the overlap.

### DIFF
--- a/book/src/interactivity/ripples.md
+++ b/book/src/interactivity/ripples.md
@@ -52,10 +52,10 @@ Combine ripples with other pressed state changes:
 
 1. **Press** - a disc appears at the click point, already at about a third of
    its final size, and starts spreading
-2. **Release** - the press happened, so the remaining expansion is *completed*
-   rather than interrupted, while the disc fades out
-3. **Or leave without releasing** - nothing was activated, so there is nothing
-   to complete: the disc just fades, quickly
+2. **Release inside** - the press happened, so the remaining expansion is
+   *completed* rather than interrupted, while the disc fades out
+3. **Release elsewhere, or drag off the button** - nothing was activated, and
+   the ripple says the same thing `on_click` does: the disc just fades, quickly
 
 Two properties are worth stating outright, because they are what makes the
 effect read as an acknowledgement rather than a hesitation:
@@ -64,9 +64,13 @@ effect read as an acknowledgement rather than a hesitation:
 - **A short press does not truncate the expansion.** A click lasts 60-150ms
   against a growth measured in hundreds, so the release finishes the growth
   instead of abandoning it half-way.
+- **No press is dimmer for being quicker.** The disc always finishes appearing
+  before it starts to leave, so a 20ms tap is as bright as a held one.
 
 The disc's centre also drifts toward the container's own centre as it grows, so
-a press near a corner settles onto the button instead of staying lopsided.
+a press near a corner settles onto the button instead of staying lopsided. Its
+final size comes from the container, not from where you pressed — every ripple
+on a given button is the same size, and a full one covers it exactly.
 
 The ripple:
 - Respects corner radius and container shape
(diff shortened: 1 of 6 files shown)


MalpenZibo:
Review turned up four defects, two visible on the first click. Fixed in `282609d`.

**A tap could be invisible.** The exit snapshotted `opacity`, which only `advance` ever raised — and events are dispatched as a batch before the frame's animation pass, so a touch tap (down and up in one batch) reached `release` with the opacity still at zero and spent its whole 375ms drawing nothing. Every click shorter than the 75ms rise was dimmer than a longer one for the same reason. The rise now always finishes before the fall begins, so a 20ms tap is as bright as a held press — Compose's rule, and it deletes the snapshot rather than patching it.

**The size came from the contact point while the centre drifted to the container's.** Those do not compose. On a 200x40 button a corner press ended at twice the radius of a centre press, and covered the box at about two thirds of the growth — the truncation this PR exists to remove, reintroduced geometrically. The radius is now the container's own half-diagonal, which is what the disc has to reach from where it ends up.

**A release that wandered off ran the "press accepted" animation** while `on_click`, ten lines below, correctly did not fire. **And dragging off the container cancelled nothing**, because only leaving the whole surface said so — the disc kept growing at full brightness on a button the pointer had left. Both are the abandoned press; both cancel now.

Also from the review: the speeds are clamped (zero divided a duration to infinity and pinned the frame loop; a negative one ran the ease backwards into a negative radius); the growth after a release is never given longer than the fade it overlaps, so no pair of speeds can leave the disc half-grown and already invisible; the geometry moved into `Ripple` beside the constant it uses, instead of being prose in one file and code in another, which also un-leaks `RIPPLE_START_RADIUS` from the public API; `SmallVec<[Ripple; 2]>` became a `Vec` — three words instead of ~144 bytes inlined into every container carrying any handler, and it was spilling at three ripples anyway; the overlay buffer is sized for overlapping ripples; and the clip is no longer installed on a frame that draws nothing.

Five new tests, each one failing before its fix: a tap released before its first frame is still seen, a click shorter than the rise still reaches full strength, the size does not depend on where the press landed, a finished ripple covers the container and not before, and an impossible speed neither stalls nor inverts the effect.

One thing from the review I did **not** act on: `advance` still reads the clock per container rather than per frame. The parameter is a test seam, and the commit message that claimed otherwise overstated it. Threading a real frame clock touches the `Widget` trait and belongs on its own.

PR #195, opened 2026-08-18
A spring keeps its momentum when it is sent somewhere else

MalpenZibo:
`AnimationState::animate_to` handed the integrator a fresh `SpringState` on
every target change — position zero, **velocity zero**. A spring interrupted
mid-flight therefore came to a dead stop and started again.

It has a symptom today, with no new feature involved: a hover that reverses
halfway snaps before it turns. What guido calls a spring has been an easing
curve shaped like one.

Compose's `Animatable` promises the opposite explicitly (continuity of value
*and* velocity across `animateTo`), and so do iOS's and SwiftUI's springs.

## What changed

The velocity crosses over into the new segment. The integrator normalises over
its own segment, so the speed is converted back into the property's own units
and then into the new segment's; the sign says whether the motion is already
heading towards the new target, and a negative one is a spring that has to turn
around — which is the whole reason to keep it.

- A value that was standing still still starts from rest.
- A timed transition has no momentum to carry, and is untouched.

`Animatable` gains `distance(from, to) -> f32` for the rescaling. Only the
ratio of two distances is ever used, so it is a magnitude and not a metric:
largest channel for a `Color`, largest side for a `Padding`, largest matrix
term for a `Transform`.

## Why it is worth more than the turn

A wobble comes free with it. Send a value out and back a moment later and the
return leg overshoots and settles — the shake a lock screen wants after a wrong
password, with no keyframes, no sequence and no timer loop. Before this, the
same two writes produced a dead stop and a slow return, which is why the shake
in the app that prompted this is currently six timed writes.

## Verified

- `a_spring_reversed_in_flight_carries_its_momentum` — the new segment starts
  with a negative velocity, i.e. still travelling away from the new target.
- `the_carried_momentum_overshoots_the_way_back` — sent back where it came
  from, the value crosses the target and returns.
- `a_spring_sent_further_the_same_way_keeps_closing`, and rest/timed cases
  pinned so the old behaviour survives where it was right.
- 399 lib tests green, clippy clean, book's springs chapter documents the
  interruption rule.

### DIFF
--- a/book/src/animations/springs.md
+++ b/book/src/animations/springs.md
@@ -157,6 +157,31 @@ fn spring_button() -> Container {
 }
 ```
 
+## Interruption
+
+A spring that is retargeted mid-flight keeps its momentum: the new segment
+starts with the speed the old one had, so a value sent back where it came from
+*turns around* rather than stopping first. That is what makes a spring feel
+physical when a hover reverses halfway, and it is also the cheapest wobble
+there is — send a value out and then back, and the overshoot on the way home
+is a shake nobody had to choreograph:
+
+```rust
+let angle = create_signal(0.0);
+
+container()
+    .transform(move || Transform::rotate_degrees(angle.get()))
+    .animate_transform(Transition::spring(SpringConfig::BOUNCY))
+
+// out, and back a moment later: the return leg crosses zero and comes back
+angle.set(2.0);
+// ...one short delay...
+angle.set(0.0);
+```
+
+Timed transitions have no momentum to keep: a retarget there starts from the
+current value at whatever the curve says, as it always has.
+
 ## API Reference
 
 ```rust
--- a/book/src/animations/springs.md
+++ b/book/src/animations/springs.md
@@ -164,23 +164,35 @@ starts with the speed the old one had, so a value sent back where it came from
 *turns around* rather than stopping first. That is what makes a spring feel
 physical when a hover reverses halfway, and it is also the cheapest wobble
 there is — send a value out and then back, and the overshoot on the way home
-is a shake nobody had to choreograph:
+is a shake nobody had to choreograph.
+
+It has to be two *different* frames, though. A press and its release are two,
+so a button shakes itself:
 
 ```rust
-let angle = create_signal(0.0);
-
-container()
-    .transform(move || Transform::rotate_degrees(angle.get()))
-    .animate_transform(Transition::spring(SpringConfig::BOUNCY))
-
-// out, and back a moment later: the return leg crosses zero and comes back
-angle.set(2.0);
-// ...one short delay...
-angle.set(0.0);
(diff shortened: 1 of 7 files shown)


MalpenZibo:
Review found the carried velocity taking its **sign** from a geometric probe — a `lerp` a hundredth of a segment ahead, compared by distance — which encodes only `sign(target - start)` and never consults the integrator's own `spring.velocity`. Every preset overshoots, so after the first crossing the spring spends long stretches travelling backwards along its segment, and in all of them the probe reported the opposite of the real motion: the new segment was kicked *away* from its target. That is the jerk this PR was written to remove, with a wrong-direction lurch added on top.

Fixed in `59b55a6` by doing the projection the probe was standing in for. `Animatable` now reports its `channels`, and the velocity a new segment inherits is the part of the old motion pointing along it:

```
v' = velocity * <old, new> / |new|²
```

One projection does three jobs:

- **The sign comes from the integrator** instead of being re-derived from geometry that cannot see it.
- **Momentum stays in its own channel.** A translation interrupted by a pure scale change carries nothing, because the two directions are orthogonal. The old scalar `distance` — an L∞ over channels of mixed units, `Transform.data` being pixels *and* unitless coefficients — would hand a 2200px/s translation to a 0.05-wide scale segment and balloon it.
- **The f32 tie cannot arise.** With `start = 1000.0` and a short segment, the probe's hundredth landed below the ulp and `closing` resolved to "reversing" on a spring that was closing.

Three more from the same review:

- **A settled spring drops its state**, so the next segment cannot inherit the residue left under the settle threshold and rescale it by however short it is. `a_settled_spring_starts_from_rest` was passing only through the degenerate-span guard `set_immediate` creates — it now lets a spring genuinely settle first, which is where the residue actually lives.
- **A transition still inside its `delay_ms` carries nothing.** Its spring is stepped only after the delay, so a velocity stored now is released at full strength against a value that has not moved since.
- **The tests no longer sleep.** `advance` reads the clock through `start_time`, so moving that back simulates exactly. The sleeping is what kept the whole bug invisible: 40ms of real sleep lands squarely inside the half of the phase space that worked, and on a loaded runner it would have started failing by itself.

New coverage: the projection primitive gets its own tests (same way keeps the speed, other way comes back negative, a negative velocity is read as the motion it is, orthogonal directions carry nothing, degenerate and near-zero segments), plus interruption past the overshoot, cross-channel isolation, a moving target staying bounded, a genuinely settled spring, and a delayed transition.

`examples/spring_momentum.rs` covers the half only a hand can check — five panels, each saying what to do to it and what should happen, since every one needs the animation interrupted at a moment only a pointer can pick.

The book snippet was not valid Rust and, as written, produced no wobble at all: the two `set` calls were top-level statements landing in one frame, so the spring only ever saw the last one. It is a press/release pair now — two frames by construction — and it says why.

Full suite green, clippy clean, book builds.

PR #196, opened 2026-08-18
A sequence can be played, not only aimed at

MalpenZibo:
Every animation in guido moves *towards* a value: a property changes and a
transition carries it there. That covers the whole of a state change and none of
a sequence — a shake, a flash, a bounce, anything whose point is the shape of
the motion rather than where it ends up.

```rust
container()
    .keyframes_transform(
        Keyframes::new(320.0)
            .at(0.0, Transform::IDENTITY)
            .at(0.15, Transform::rotate_degrees(2.0))
            .at(0.40, Transform::rotate_degrees(-1.6))
            .at(0.65, Transform::rotate_degrees(0.9))
            .at(1.0, Transform::IDENTITY),
        rejections,
    )
```

## The type

`Keyframes<T>` is stops on a run: a fraction, a value, and the easing of the
segment *leaving* it — CSS's rule for a timing function written inside a
keyframe. Stops written out of order are sorted, the ends hold rather than
extrapolate, `repeat(n)` plays the run again, and `value_at` returns `None` once
it is over, which is how the property gets handed back.

## What plays it

A signal, and every change to it plays the sequence once. A count and not a
flag, for the reason the app that prompted this had already discovered: two
refusals in a row are two events, and a signal that stays equal notifies nobody.
SwiftUI's keyframe animator takes its trigger the same way.

Nothing plays on the first frame — the container remembers what the signal held
when it was built — and playing again mid-run restarts it, because half a shake
is not what a second refusal means.

## Replace, not compose

While a timeline runs it **replaces** the declared value, and hands it back at
the end — to whatever is declared *then*, including a value that changed while
it was playing. That is the cascade rule CSS gives an animation over a normal
declaration, and `animation-composition`'s default. It is what lets a card
declare a hover transform it is simply not showing for those 320ms:

```rust
.when_hovered(|s| s.transform(Transform::scale(1.03)))
.animate_transform(Transition::spring(SpringConfig::SNAPPY))
.keyframes_transform(shake(), rejections)
```

Composition (CSS's `add`/`accumulate`) is deliberately not here. It is only
well-defined for some types — concatenation for transforms, addition for
scalars, nothing sensible for colours — and nothing has asked for it yet.

## Scope

Transform only for now. The machinery in `AnimationState` is generic, so each
further property is one builder — but a colour that flashes is usually a state
layer, and I would rather add them when something actually wants one.

## Verified

- Unit tests on the type: passing through the stops, sorting, the ends holding,
  repeats, per-segment easing, and an empty timeline playing nothing.
- On `AnimationState`: a timeline plays and then hands the property back to a
  declared value that changed meanwhile; playing again restarts it; a property
  with no timeline cannot be played.
- On the container: a change to the trigger queues an `Animation` job, i.e. the
  container is subscribed to the signal that plays it.
- `cargo run --example keyframes_example` — a shake and a nod from one trigger,
  over a hover that stays declared throughout.
- 405 lib tests green, clippy clean, book chapter added.

Depends on nothing else, but reads best after #195 (springs keeping their
momentum), which is the reason the "when not to use them" section can say that
out-and-back no longer needs a timeline.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -42,6 +42,7 @@
     - [Transitions](animations/transitions.md)
     - [Timing Functions](animations/timing.md)
     - [Spring Physics](animations/springs.md)
+    - [Keyframes](animations/keyframes.md)
     - [Animatable Properties](animations/properties.md)
 
 # Transforms
--- a/book/src/animations/keyframes.md
+++ b/book/src/animations/keyframes.md
@@ -68,8 +68,28 @@ container()
     .keyframes_transform(shake(), rejections)
 ```
 
-The hover is still declared while the shake runs; it simply is not being drawn,
-and it has the property back the moment the sequence ends.
+The hover is still declared while the shake runs; it simply is not being drawn.
+When the sequence ends the property is handed back **through its declared
+transition**, from wherever the sequence left it — so a pointer that arrived
+mid-shake gets its spring, rather than the card jumping to the hovered size on
+the sequence's last frame. A callback on that transition fires as it normally
+would, for the same reason.
+
+The builders do not care which order you write them in: declaring a transition
+after a sequence keeps the sequence, and the other way round too.
+
+## What a segment cannot be eased with
+
+A spring. A keyframe segment is a fixed slice of a fixed duration, which is
+exactly what a spring has not got — it settles when it settles. Passing one to
+`at_with` substitutes `EaseInOut` and says so in debug builds, rather than
+quietly playing the segment as a straight line.
+
+```rust
+Keyframes::new(320.0)
+    .at_with(0.0, Transform::IDENTITY, TimingFunction::EaseOut)
+    .at(1.0, Transform::rotate_degrees(2.0))
+```
 
 ## When not to use them
 
--- a/book/src/animations/keyframes.md
+++ b/book/src/animations/keyframes.md
@@ -1,0 +1,83 @@
+# Keyframes
+
+Every other animation in guido moves *towards* something: a property gets a new
+value and a transition carries it there. That is the whole of what a state
+change looks like, and none of what a sequence looks like — a shake, a flash, a
+bounce, anything that has to pass through somewhere on its way and end where it
+started.
+
+A timeline is the other shape. It has no target: it plays.
+
+```rust
+container()
+    .keyframes_transform(
+        Keyframes::new(320.0)
+            .at(0.0, Transform::IDENTITY)
+            .at(0.15, Transform::rotate_degrees(2.0))
+            .at(0.40, Transform::rotate_degrees(-1.6))
+            .at(0.65, Transform::rotate_degrees(0.9))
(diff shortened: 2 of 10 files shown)


MalpenZibo:
Merged main in (the spring work and this both rewrote the tail of `animations.rs`), then took the review's findings.

**A reachable panic.** `f32::clamp` passes NaN through, so a stop whose offset was computed — `i as f32 / count` with an empty count is all it takes — was stored as NaN, compared false against every early return in `sample`, and sent the search to `stops[0 - 1]`. Underflow panic in the render loop on the first frame of the play. The offset is absorbed at the door now, and the index has a floor so the case is closed structurally.

**The order the builders were written in decided whether the sequence existed.** `animate_transform` builds a fresh `AnimationState`, so written after `keyframes_transform` it dropped the timeline — while the trigger went on firing, the job went on being queued, and nothing played. `adopt_timeline_of` carries it across, and there is a test that fails without it.

**The end of a run cancelled the declared transition.** Forcing `progress = 1.0` and returning `target` made the case the feature is sold on jump: a card shaking, a pointer arriving mid-shake, and the hover landing instantly with no spring. The transition was suspended for the duration, not cancelled — the hand-back now starts a fresh segment from where the sequence left the value and lets the declared transition run it home. Same cause, same fix, for `on_complete`, which was never reaching its completion edge.

**A `playing` could outlive its sequence**, leaving `is_animating` true for good — a surface asking for a frame every vsync with nothing to draw. The two are read together now.

**Altitude.** The timeline and its trigger move into one boxed `Timeline<T>`, so the other eight animation slots in `ContainerAnims` stop paying ~48 bytes each for a field they cannot set. And `wants_play`/`take_play` make asking and committing one comparison in one place, instead of a read in `resync_animation_targets` and a write in `advance_animations` that nothing kept in step — which is also what makes a future `keyframes_background` just a builder.

**A spring passed to `at_with`** played the segment dead linear, since `TimingFunction::evaluate` returns `t` unchanged for one. A segment of fixed duration is what a spring has not got, so it is replaced with `EaseInOut` and said out loud in debug.

**`Keyframes` was not in the prelude**, so the chapter's snippets did not compile as written — the example importing it by hand was the tell.

New coverage: the NaN offset, the spring easing, the builder order (red before the fix — verified), the hand-back animating rather than snapping, `on_complete` surviving a sequence, a play without a sequence not pinning the loop, the trigger being taken once per change, and a container-level test that asserts the transform *moves* rather than that a job was queued — which is what the old one checked, and why it passed either way.

The timeline tests no longer sleep: they step the run by moving its start back, the same way main's spring tests do.

The book chapter now says what "the property comes back" means, that the order of the builders does not matter, and what a segment cannot be eased with.

Full suite green, clippy clean, book builds.

One thing I have **not** done: split the original commit. CLAUDE.md's atomic-commit rule wanted the type, the playback machinery, the widget API and the docs as separate increments; the branch landed them as one. My changes are in three commits along those lines, but the base commit is what it is — say the word if you want the whole thing restructured before merge.

Issue #209, opened 2026-08-23
Manual test round: paths left unreached for want of hardware


MalpenZibo:
A manual pass over the examples (during #197) turned up eleven bugs — #198
through #208 — all of them older than that branch. It also left a few paths
unreached, for want of hardware rather than for want of looking. Recording them
so they are retested deliberately rather than by accident.

### Multi-output hotplug

`examples/multi_output.rs` spawns a bar per monitor from the reactive `outputs()`
list, each pinned with `.output(id)` and reserving with `ExclusiveZone::Auto`.
With one screen it exercises `spawn_surface` at runtime, `Auto` on a dynamically
created surface, and `surface_output()` — all of which behaved.

What it cannot reach with one screen is the reason `outputs()` is reactive at
all: **plugging a monitor in must add a bar, unplugging must remove one**, along
with the teardown of that surface and its reservation. Needs a second display.

Worth reaching, because a surface disappearing under the compositor's feet is
the shape of failure that has been expensive elsewhere — see #208, where a
nested popup's dismissal signal is owned by the thing that dies with it.

### Session lock

`ext-session-lock-v1`, one lock surface per output with hotplug handling
(`examples/simple_lock.rs`). Not run: it locks the screen, which is disruptive
mid-session, and its own hotplug half needs the second monitor too.

### Touch input

`wl_touch`, where the first finger drives the pointer pipeline — tap as click,
with the hover and pressed state layers following. No touchscreen on the machine
the pass ran on.

### How to close them

Each is a case of "the example exists and is correct as far as it goes". The
round's lesson was that the untested paths are where the bugs were: three of the
eleven were found by examples written that day (#202, #203, #208), and one of
those kills the application. None of these three has an equivalent to lean on.

MalpenZibo:
## Two of the three reached, on hardware, 2026-09-21

An external monitor (LG 4K on DP-2 at scale 1.5, beside eDP-1 at scale 2) closed
the first two sections. Both found something.

### Multi-output hotplug — reached

`examples/multi_output.rs`, DP-2 unplugged and replugged physically.

Plugging in and unplugging do add and remove a bar, and `surface_output()` was
correct throughout. But unplugging also leaves a **phantom output** behind, and
with it a bar labelled after the monitor that just left, sitting on the
remaining screen and reserving a second exclusive zone. `outputs()` went on
reporting 2 connected monitors with one plugged in, for as long as nothing else
happened — 66 seconds in that run, ending only because the monitor came back.
[...]

PR #197, opened 2026-08-19
Public API audit: predictable reactivity, one spelling per idea, two preludes

MalpenZibo:
An audit of the whole public surface, read against a real application built on
top of it (allock). What it turned up was not missing features but
**inconsistency**: the same idea spelled two ways, or reactive in one place and
constant in the one next to it, so nothing about the API could be predicted
from the rest of it.

Ten atomic commits, each reviewable on its own.

## Reactivity now has a rule

Which container properties took a signal was arbitrary: `background` did and
`gradient` did not, `corner_radius` did and `backdrop_blur` did not, `visible`
did and `overflow` did not. Every gap forced the caller to branch in Rust and
rebuild the widget where the neighbouring property would simply have followed a
signal.

The sharpest case is `backdrop_blur`, which carried **two different contracts
under one name**: on `Text` a reactive radius where `0.0` means off, on
`Container` a constant.

```rust
container()
    .gradient(move || palette.get().header())
    .backdrop_blur(move || if frosted.get() { 24.0 } else { 0.0 })   // 0.0 = off
    .overflow(move || if collapsed.get() { Overflow::Hidden } else { Overflow::Visible })
```

The rule is now stated in `docs/STYLING.md` and the book: **anything that
survives to paint takes a signal; anything structural does not** (`layout`,
`scrollable`, `scrollbar`, `scrollbar_visibility`, `control`, `animate_*`).

`overflow` decides whether a box may shrink below its content, so it reaches
layout as well as paint: read inside the layout tracking scope alongside the
lengths it belongs with, and threaded through `BoxLengths` to the axis
resolver.

Also here, from the same audit: `border_width` and `border_color` as separate
setters (the state layer has had them all along, so a layer that only recolours
the frame no longer restates its width), and `gradient_diagonal` /
`animate_elevation`, both of which the book already documented.

## One spelling per idea

| | before | after |
|---|---|---|
| two-value padding | `Padding::symmetric(h, v)` **vs** `[v, h]` — opposite orders, both `f32`, no warning | only `[v, h]`; `horizontal_total()` says it is a sum; `with_top()` no longer collides with the field |
| effects | `create_effect(..)` returned a guard that disposed itself *unless* an owner had claimed it | returns `()`; an effect's lifetime is its scope's |
[...]

PR #223, opened 2026-08-24
Queueing work for the loop is the same call as waking it

### LINKED ISSUE #202: The pre-block assertion fires on healthy states: it checks the queues but not whether a wakeup is armed
`App::run` asserts, before blocking with no timeout, that nothing is queued that
nobody owes it a wakeup for:

```rust
debug_assert!(
    queued_but_unwoken().is_none(),
    "the loop is about to block indefinitely with {} still queued — \
     whatever produced it owes a wakeup. ..."
);
```

`queued_but_unwoken` checks seven queues for emptiness — and nothing else. But
"unwoken" is the other half of the question, and every producer *does* wake the
loop: `queue_bg_write` sends on the ingress channel, `push_surface_command`
calls `jobs::wake_loop()`. Both wakeups survive until the loop reads them, so a
queue left non-empty while one is outstanding is not the state this assertion
looks for — it is the ordinary one, work queued after this frame's drain points,
riding the next iteration.

The result is a `debug_assert` that fires on healthy states. Three distinct
windows, all reproduced in minutes with a background task writing a signal every
16ms and an effect calling `SurfaceHandle::set_margin`:

**1. Background writes, queued after the frame's flush.**
`flush_bg_writes()` runs late in the iteration; the pre-block check is at the top
of the next one. A write landing in between leaves the queue non-empty with the
ingress doorbell already sent and not yet delivered. Blocking is safe — calloop
delivers the message before the loop can block again — and the assertion fires
anyway.

```
panicked: the loop is about to block indefinitely with background signal writes
still queued
```

**2. Surface commands, queued from a main-thread effect.**
`push_surface_command` calls `wake_loop()`, which pings; the ping stays readable
until the next dispatch consumes it, so the loop cannot block. `take_wake_request()`
then clears `WAKE_REQUESTED` later in the same iteration, so by the pre-block
check neither the flag nor the queue-emptiness test can see that a ping is armed.

```
panicked: the loop is about to block indefinitely with surface commands still
queued
```

**3. The window inside `queue_bg_write` itself.**

```rust
if let Ok(mut q) = WRITE_QUEUE.lock() {
    q.push((epoch, Box::new(f)));      // the queue is non-empty from here
}
crate::ingress::notify(IngressMessage::BgWritesQueued);   // armed only here
```

Between the push and the notify the queue is observably non-empty with nothing
yet owed for it. Narrow, but a loop spinning through iterations (moving the
mouse over the surface is enough) lands in it within seconds.

### What a fix has to decide

Teaching the assertion to ask "is a wakeup armed?" is the direction, and it is
not one line — I tried it twice while walking the examples and it was incomplete
both times, which is why this is an issue rather than a patch. The pieces:

- **Two messengers, both of which must be askable.** A calloop channel send and a
  wakeup ping. `PING_SENT` already tracks the second (it is cleared once per
  wakeup by `mark_loop_awake`), so `wake_ping_armed()` is nearly free; the first
  needs a counter incremented in `ingress::notify`.
- **When the counter falls.** Decrementing on *delivery* is wrong: the doorbell
  is consumed at dispatch while the write queue is drained later in the same
  iteration. It has to fall at the drain point, which means arming is a property
  of the queue rather than of the message.
- **Window 3 closes only by arming before the push**, so the arm and the send are
  two steps rather than one.
- **`reset_ingress` / `reset_jobs`** must clear whatever is added, or an `App`
  restart starts armed.

An alternative worth weighing: drop the emptiness test for queues whose producer
provably always wakes, and keep the assertion only for those where a producer
could forget. That is a smaller change and admits what the assertion can actually
prove.

### Not from #197

`src/ingress.rs`, `src/jobs.rs` and the assertion itself are unchanged between
`main` and the API-audit branch. Found while walking the examples for
regressions, with `examples/surface_runtime.rs` (added in that branch) driving
the four `SurfaceHandle` setters at frame rate.

Debug builds only — `debug_assert!` — but that is where the examples run.

MalpenZibo:
Closes #202.

## The report

The loop's pre-block `debug_assert!` named seven queues and panicked if one was non-empty. It fired on healthy states: a background write landing after `flush_bg_writes()`, a surface command pushed by an effect whose `WAKE_REQUESTED` was cleared by `take_wake_request()` later in the same iteration, and the instant inside `queue_bg_write` between the push and the notify. Reproducible in seconds — a debug build with a `create_task` writing a signal, which is the documented way to do background work.

## Where it landed

The first answer was to make the check ask the other half of the question: *is a wakeup armed?* That needed a coalescing flag for the ping and a count for messages in flight inside calloop's channel — about 150 lines and a new rule (`arm` before you queue) that every future producer would have to learn.

That was policing a rule the type system can make unbreakable, so the check is gone and the rule is structural instead. `src/deferred.rs`: a deferred queue and its wakeup are one object. `DeferredQueue::push` and `DeferredSlot::set` **are** the wakeup, the cell is private, and there is no way to reach it that does not ask for the pass that empties it. Six producers moved onto them — owner disposals, surface commands, cursor, clipboard, primary selection, and the session lock's request.

Three stay outside, each for a reason the module records: background writes wake through the ingress channel and there is exactly one of them; widget jobs carry their own machinery and wake from inside `request_job`; a parked focus request has the opposite invariant, since waiting for a widget that does not exist yet is its resting state.

Net: **−350 lines** against the branch's own high-water mark, and the state the assertion existed to report is now unrepresentable rather than reported.

## Bugs found and fixed on the way

- **A drain that only ran when a surface had a frame.** `sync_platform_state` sat inside the per-surface pass, so an application with no configured surface could queue a copy or a cursor shape that nothing would ever take. It runs once per iteration from the loop body now — where what it carries belongs anyway, since it is the seat's, not any one surface's.
- **A flag that said "requested" and was read as "readable".** `wake_loop` raised `PING_SENT` before trying the handle, so a second caller could coalesce against a ping that was never written. It is raised under the ping's lock, with the handle in hand.
- **`quit_app` could be coalesced away.** The loop reads the exit flag once per pass, so an exit request cannot be left to somebody else's ping. It writes its own, unconditionally.
- **A clipboard read could land in the next session.** The reader thread now binds its sender to the loop that was running when the read started; generations restart with the `WaylandState`, so a late result would otherwise match a number that means something else.
- **Owner disposals survived `App::drop`.** Owner ids restart with the arena, so a disposal queued on the way out named a live owner in the next App.
- **The wakeup mechanisms were installed too late** — after `create_wayland_app` and the configure loop, and after `setup`, which is where an app's first `create_task` runs.
- **The session lock asked for two things at once.** `lock_session` and `unlock_session` set independent booleans read three steps apart, so a lock could go out to the compositor and be undone in the same iteration. One slot, last answer wins.
- **`cargo test --release` had never compiled** (`report_count` gated `all(debug_assertions, test)`, its callers `cfg(test)` — since #148). With the target building again it immediately caught two tests of this branch asserting on a debug-only count.

## Verification

- Full suite green in debug and release; `cargo clippy --all-targets --all-features -- -D warnings` clean in debug, `--lib` clean in release.
- The reported failure, live on niri: `main` panics within seconds of the surface coming up, this branch survives 15 s and 60 s runs, exit 0. A second repro with no surface at all — an effect calling `set_cursor` — panics on the pre-move code and is clean here.
- Session lock exercised on the compositor: two full cycles, both unlocked by the password path, lock surfaces created from the factory carried inside the request and torn down, no warnings.
- Performance against `main`, release, `/proc` sampling: idle is **0 ms of CPU over 45 s** on both, and a background task writing a signal once a second costs the same CPU (0.33–0.35%) with 3–4% fewer wakeups on this branch.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -319,6 +319,12 @@ either carry their payload or act as doorbells for data queued elsewhere
 flush point). Never call `jobs::wake_loop()` directly from a background
 thread as the *only* wakeup for queued work.
 
+Messages are counted while in flight, because a wakeup living inside
+calloop's channel is invisible from the sending side and the loop's pre-block
+check has to be able to ask about it. `ingress::arm()` puts the count up and
+hands back the message to send; the loop's channel callback takes it back down
+when it reads one.
+
 **Main thread → `jobs::wake_loop()`.**
 The frame-request ping is coalesced per loop iteration through a dedicated
 `PING_SENT` flag cleared once per wakeup (`mark_loop_awake`, right after
@@ -350,17 +356,36 @@ A focused input is the resting state of a lock screen, so that ran all night.
 
 **The contract is checked, not just written down.** Debug builds assert it at
 the one moment where breaking it is fatal: `queued_but_unwoken()` in
-`src/lib.rs` names every queue the loop drains, and nothing may be
-outstanding when the loop is about to block indefinitely. The reasoning is
-that each queue is drained unconditionally once per iteration, so anything
-still queued was produced *after* its own drain — and the only thing that
-brings the loop back is a frame request, which would have kept it from
-blocking. A non-empty queue there means nobody asked to be woken.
+`src/lib.rs`, run just before the loop blocks with no timeout.
 
-So a new queue needs one line in `queued_but_unwoken()` alongside its drain.
-Forget the wakeup and the next idle moment panics with the queue's name,
-instead of the app going quietly deaf until an unrelated compositor event
-happens along. Two of the bugs documented above were exactly that.
+It asks two questions, and needs both. `queued_work()` names every queue the
+loop drains and answers *what* is waiting. `wakeup_armed()` answers whether
+anything is coming to drain it — a ping still readable (`jobs::ping_in_flight`,
+the coalescing flag, cleared right after the dispatch that consumed it), a wake
+request still set, or an ingress message sent and not yet read
+(`ingress::in_flight`). Only a queue with none of those is reported.
+
+The second question is not optional, because a non-empty queue at that point
+is the ordinary case rather than the bug: the drains sit in the middle of the
+iteration and effects, event handlers and background threads all run after
+them, so work riding the next pass is what a working application looks like.
+An earlier version of this check asked the first question alone and panicked
+on healthy states — a background write landing after `flush_bg_writes()`, a
+surface command pushed by an effect whose `WAKE_REQUESTED` was then cleared by
+`take_wake_request()` later in the same iteration.
+
+The order of the two is part of the contract: **a producer arms before it
+queues**, so a queue seen non-empty already has its wakeup armed, and reading
+the arming first can miss one that lands a moment later on another thread.
+That is why `queue_bg_write` calls `ingress::arm()` before pushing and sends
+after — between the push and the send there is an instant where the queue is
+observably non-empty, and the main thread reads it.
+
+So a new queue needs one line in `queued_work()` alongside its drain, and its
+producer needs a wakeup through one of the two mechanisms above. Forget the
+wakeup and the next idle moment panics with the queue's name, instead of the
+app going quietly deaf until an unrelated compositor event happens along. Two
+of the bugs documented above were exactly that.
 
 ## Widget Trait
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -320,11 +320,12 @@ flush point). Never call `jobs::wake_loop()` directly from a background
 thread as the *only* wakeup for queued work.
 
 There are two ways in, and `ingress::sender` is private so there is no third.
-`notify()` resolves the sender at send time, for a producer whose payload is
-ready now. `IngressSender` is taken *before* the work starts and binds the send
-to the loop that was running then — a selection read has three seconds to
-finish, and a result that arrives after a restart would otherwise land in a
-loop whose generation counters have started over.
+`notify()` sends in the same call that queued the work, for a producer whose
+payload is ready then and there. `IngressSender` is taken *before* the work
+starts and binds the send to the loop that was running at that moment — a
+selection read has three seconds to finish, and a result that arrives after a
+restart would otherwise land in a loop whose generation counters have started
+over. Both hand the wakeup to the ping if the receiver has gone.
 
 **Main thread → `jobs::wake_loop()`.**
 The frame-request ping is coalesced per loop iteration through a dedicated
@@ -372,7 +373,7 @@ a coalescing flag for the ping and a count for messages in flight in the
 calloop channel. That is a lot of apparatus to verify at runtime a rule the
 type can make unbreakable.
 
-Three producers stay outside `deferred`, each for a reason worth knowing:
+Four producers stay outside `deferred`, each for a reason worth knowing:
 
 - **background writes** wake through the ingress channel rather than the ping,
   and there is exactly one of them (`queue_bg_write`), so the pairing is one
@@ -381,8 +382,14 @@ Three producers stay outside `deferred`, each for a reason worth knowing:
   per-surface lanes — and wake from inside `request_job`;
 - **a parked focus request** is the opposite invariant: it waits for a widget
   that may not be laid out for many frames, so *still full* is its resting
-  state, not a failure. It is applied once per iteration after the render
-  pass, where the tree is laid out.
+  state, not a failure. It is applied at the end of `layout_pass`, after the
+  tree has resolved and before the paint that shows it;
+- **the session lock's request flags** (`lock_session` / `unlock_session`) are
+  still a flag and a `wake_loop()` written as two statements. They belong in
+  `deferred` by shape — `process_session_lock` drains them unconditionally
+  once per iteration — and are not there yet because the request carries a
+  widget factory into a state machine with its own guards, and the path cannot
+  be exercised without locking the screen (see #209).
 
 Everything in `deferred` is drained unconditionally, once per iteration, in
 the loop body. That is what makes "the loop will get to it on the next pass"
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -321,9 +321,20 @@ thread as the *only* wakeup for queued work.
 
 Messages are counted while in flight, because a wakeup living inside
 calloop's channel is invisible from the sending side and the loop's pre-block
-check has to be able to ask about it. `ingress::arm()` puts the count up and
-hands back the message to send; the loop's channel callback takes it back down
-when it reads one.
+check has to be able to ask about it. There are two entry points, and the
+difference is what the message stands for:
+
+- `ingress::arm()` raises the count *before* the work is queued and hands back
+  the message to send after — for a doorbell like `BgWritesQueued`, whose
+  payload sits in a queue the loop can already see;
+- `IngressSender::send` raises it around the send — for a message that carries
+  its own payload, which leaves no queue behind and so has nothing to arm
+  before. Its sender is bound to the loop that was running when the read
+  started, so a slow result cannot land in a later session's loop.
+
+The loop's channel callback takes the count back down when it reads a message.
+Nothing else may send: `ingress::sender` is private precisely so that every
+message the callback decrements for was counted up by one of these two.
 
 **Main thread → `jobs::wake_loop()`.**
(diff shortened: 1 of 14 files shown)


MalpenZibo:
## Verified on a live compositor (niri, debug build)

Same example both sides — a `create_task` writing a signal every 16 ms, plus an effect on that signal calling `surface_handle(id).set_margin(..)`. Nothing exotic: window 1 and window 2 of the report, as an application would write them.

**On `main` (`eeb9a40`)** — panics within seconds of the surface coming up:

```
thread 'main' panicked at src/lib.rs:1535:17:
the loop is about to block indefinitely with background signal writes still queued —
whatever produced it owes a wakeup. See the Event Loop Wakeup Contract in docs/ARCHITECTURE.md.
```

**On this branch** — 15 s and 60 s runs, both clean, exit 0, margins cycling and the compositor reconfiguring the surface throughout.

The example used (not committed — the regression lives in the unit tests):

```rust
#[tokio::main]
async fn main() {
    App::new().run(move |app| {
        let tick = create_signal(0u64);
        let tick_w = tick.writer();

        // Window 1: a background write landing wherever it lands in the
        // loop's iteration, including after the flush point.
        create_task(move |ctx| async move {
            let mut n = 0u64;
            while ctx.is_running() {
                n += 1;
                tick_w.set(n);
                tokio::time::sleep(Duration::from_millis(16)).await;
            }
        });

        let id = app.add_surface(
            SurfaceConfig::new()
                .height(40)
                .anchor(Anchor::TOP | Anchor::LEFT | Anchor::RIGHT)
                .layer(Layer::Top)
                .namespace("wakeup-assert-repro")
                .background_color(Color::rgb(0.1, 0.1, 0.15)),
            move || container().padding(8.0).child(text(move || format!("tick {}", tick.get()))),
        );

        // Window 2: an effect running at the flush point pushes a surface
        // command, after that queue's drain.
        create_effect(move || {
            let n = tick.get();
            surface_handle(id).set_margin(if n % 2 == 0 { 0 } else { 8 });
        });
    });
}
```

MalpenZibo:
Tutti e quattro i 🟠 sistemati, più i 🟡. Due commit: `6dc95c3` per la review, `4e23577` per una rottura del release build che il primo introduceva.

## 🟠 La drain che girava solo con una surface

Verificato: `sync_platform_state` stava dentro il passaggio per-surface, oltre l'early return di `open_frame`. Un'app che copia o setta un cursore prima che esista una surface configurata lasciava quelle code piene senza nessun passaggio che le drenasse, e il check chiamava lost wakeup una cosa che non lo era.

Provato dal vivo con un esempio senza nessuna surface (un effect che chiama `set_cursor`, cinque secondi, esce):

```
thread 'main' panicked at src/lib.rs:1589:21:
the loop is about to block indefinitely with a cursor change still queued and no
wakeup armed for it — whatever produced it woke nobody.
```

Ora `sync_platform_state` gira una volta per iterazione dal corpo del loop, dopo il passaggio di render (così quello che l'event handling di questa iterazione produce esce in questa iterazione) e fuori dall'`if let Some(renderer)` (così esce anche senza nessuna surface). È dove quello che trasporta appartiene comunque: il seat, non una surface — e per-surface significava solo che la prima surface vinceva la take.

## 🟠 `ping_in_flight` diceva "richiesto"

Esatto, e con la fix precedente il buco era peggiore di come lo descrivi: `wake_loop` alzava `PING_SENT` prima di provare l'handle, quindi il check avrebbe benedetto proprio lo stato in cui il loop stava davvero per assordarsi. Il flag si alza ora sul ramo che scrive l'eventfd e si riabbassa quando non c'era niente su cui scrivere.

Hai ragione anche sulla vacuità del test: nel binario dei test nessun `WAKEUP_PING` è installato, quindi osservava solo sé stesso. Ora installa un ping calloop vero e asserisce che il dispatch si sveglia davvero — e la direzione opposta ("nessun handle ⇒ flag giù"), che è la regressione. Con la vecchia `wake_loop` quel test fallisce.

## 🟠 `Armed::notify` disarmava prima di svegliare

Righe invertite. Una nota onesta: l'ordine in sé non è osservabile da un thread solo, quindi il test nuovo (`giving_up_on_the_channel_hands_the_wakeup_to_the_ping`) copre le due postcondizioni — il loop svegliato e il conto restituito — non la sequenza. Prende un fallback che si dimentica di svegliare; non prenderebbe un fallback che sveglia dopo. L'argomento per l'ordine resta nel commento.

## 🟠 La clipboard di una sessione nella successiva

Regressione mia, confermata: le generation ripartono con `WaylandState`, quindi un contenuto vecchio poteva passare il check contro un numero uguale di un'altra sessione. Il sender torna a essere preso allo spawn, dentro `IngressSender` — un wrapper che tiene il conteggio attorno al send, così legare il send a un loop preciso non è una scorciatoia per saltare il check.

## 🟡 Test

- `set_cursor` alterna le forme: avevi ragione, l'early return su forma uguale rendeva i 64 tentativi 63 no-op.
- Un lock solo (`jobs::wakeup_test_lock`) serializza i test che scrivono i flag globali, invece di 64 retry a testa. I retry restano dove servono davvero: i test di `memo` chiamano `reset_jobs()` senza prendere il lock, e quelli non li serializza nessuno.
- Aggiunto il test sul ramo di fallback di `notify`, che è quello che il doc comment descriveva e che `Drop` non copriva.
- `queued_work() == Some("a cursor change")` l'ho lasciato. Oggi non è racy — nessun test nel binario accoda scritture di background (`WriteSignal::set` prende il ramo same-thread quando il signal vive nel thread del test, e non c'è nessun `thread::spawn` in `src/` o `tests/`) — ma la fragilità che descrivi è reale: se un giorno qualcuno scrive quel test, questo fallisce con `Some("background signal writes")`. Fallisce rumorosamente, non silenziosamente, e mi è sembrato il compromesso giusto rispetto ad aggiungere cerimonia attorno a una sonda globale.

## 🟡 Resto

Tutto accolto: singola valutazione nel panic (era un difetto che precedeva questa PR e stava per sopravviverle), `wake_request_pending` fuori da `wakeup_armed` con il perché scritto, `AcqRel` sul conteggio, lista in ordine di drain reale, `delivered`/`disarm`/`release_one` collassate.

Sul costo in release: il `#[cfg(debug_assertions)]` sul solo contatore **rompeva il build** — `in_flight()` spariva mentre `wakeup_armed` continuava a nominarlo. Ora il conteggio è un modulo a due facce (debug: l'atomico; release: no-op e `false`), quindi un solo posto dice "debug only" e niente a valle ha bisogno di `cfg`. `cargo clippy --release --lib --all-features -- -D warnings` pulito.

Nota a margine trovata così: `--all-targets` in release non compila, e non per questa PR — `reactive::diagnostics::report_count` è `cfg(all(debug_assertions, test))` mentre i suoi chiamanti sono `cfg(test)`, quindi il target dei lib test non ha mai costruito con le debug assertion spente (d55ec42, #197). Non toccato qui.

## Verifiche

Ogni fix fallisce il proprio test senza di sé, controllato uno per uno. Suite completa verde, e i due repro rieseguiti dal vivo su niri: quello con surface (write di background + comandi di surface, 20 s) e quello senza surface (5 s), entrambi exit 0.

MalpenZibo:
## Session lock verified on a compositor (niri)

`examples/simple_lock.rs`, two full cycles, both unlocked by the password path (`on_submit` → `unlock_session()` from inside event dispatch) rather than by the example's 30-second safety timer:

```
20:40:42  Session lock requested
20:40:43  Session lock granted by compositor
20:40:43  Created lock surface SurfaceId(1) on output OutputId(0)   ← 1440x960, first frame presented
20:40:46  Session unlocked
20:40:46  Destroyed surface SurfaceId(1)

20:41:27  Session lock requested
20:41:28  Session lock granted by compositor
20:41:28  Created lock surface SurfaceId(2) on output OutputId(0)
20:41:33  Session unlocked
20:41:33  Destroyed surface SurfaceId(2)
```

No warnings, no errors, and in particular no `Session locked but no lock widget factory is set` — which is the failure the change could have introduced.

What the two cycles exercise, given the request now carries its factory:

- the request travelling through `DeferredSlot<LockRequest>` into step 1 of the state machine, `Unlocked → Locking → Locked`;
- the lock surface built from the factory *inside* the request rather than one the caller installed into `LockData`;
- the second cycle replacing a factory that a previous cycle had already installed — the case that distinguishes the new ownership from the old;
- `unlock_session()` from an event handler, applied at step 3 after the compositor's events, with the surface torn down.

Still unreached, for want of hardware, and unchanged by this PR: hotplug while locked (needs a second monitor) and the refusal path (`Finished` without our unlock, i.e. another lock client already active). Both recorded in #209.

PR #254, opened 2026-08-26
[throwaway] Evidence run for #253 — do not merge

### LINKED ISSUE #253: The mutation job builds sixty examples for every mutant, and fills the runner's disk
### Observed

The `Mutation testing` job on #252 ended with:

```
ERROR Worker thread failed: write to list file
    No space left on device (os error 28)
Error: write to list file
```

It had found 16 mutants, tested one of them in `240s build + 10s test`, and run
for 27m41s before the disk gave out. It reports rather than blocks, so nothing
was gated — but a sensor that dies halfway and a sensor that finds nothing look
identical from the outside, which is the failure this repository has already had
once, when a wrong ICD path made every golden skip and report green.

The cause is not the runner. `.github/workflows/ci.yml` runs:

```
cargo mutants --in-diff /tmp/pull-request.diff -j 4 --no-shuffle
```

With no cargo arguments, that runs a plain `cargo test`, which builds **all 60
example binaries** — each statically linked against wgpu and carrying full debug
info. Measured on a clean worktree:

| build | command | `target/` | of which `examples/` | time |
| --- | --- | --- | --- | --- |
| what CI runs today | `cargo test` (CI passes no features; measured with `--all-features`, and both rows the same way, so the ratio holds) | **26 GB** | 18 GB | 84 s |
| target selection | `… --lib --tests` | 6.8 GB | 0 | 58 s |
| and without debug info | `… CARGO_PROFILE_{DEV,TEST}_DEBUG=0` | **1.4 GB** | 0 | 47 s |

Each of the `-j N` build directories compiles that from scratch. Nothing is
shared between them and `--copy-target` is off by default, so there is no
incremental reuse to soften it — which the failing log corroborates: a 184 s
baseline build, then 240 s for the very next mutant. Four trees at 26 GB do not
fit a hosted runner, and every pull request touching a non-trivial Rust file
will hit it.

The 240s build is the same cause: 96% of each mutant's cost is compiling
examples that are never run. `cargo test` builds them only to check they still
compile, and the `Test` job already does that once.

`AGENTS.md` documents the by-hand invocation as `cargo mutants -f '…' -C --lib
-j 8` — with the flag. CI is the one place that lost it.

### Expected

The job completes. Its build carries the tests and nothing else, and its
footprint fits the runner with room to spare: 1.4 GB per build directory, about
7 GB for a baseline plus four jobs.

### Acceptance

The `Mutation testing` job runs to completion on a pull request and reports its
caught/missed counts, where today it exits 1 on `os error 28`. Concretely, on
the pull request that makes this change:

1. the job's conclusion is success, not failure;
2. its log ends with a mutants summary rather than a disk error;
3. no `target/debug/examples` is built — visible in the build output, and
   already measured above.

This is the one kind of change whose proof is the CI run itself; there is no
unit test for a workflow file. Note that the pull request making this change
cannot be its own proof: `cargo mutants --in-diff` on a diff containing no Rust
file prints `Diff changes no Rust source files` and exits 0 without building
anything, so the job would go green having exercised nothing. The evidence has
to come from a run whose diff contains Rust.

### Non-goals

- Making the job block. It reports, and whether the ratchet is worth the
  friction is a separate decision.
- Freeing disk with an action like `jlumbroso/free-disk-space`. That treats the
  symptom; 130 GB does not fit whatever is freed.
- Reducing `-j`. It makes an already-slow job slower and does not address the
  cause.
- The by-hand invocation in `AGENTS.md`, which is correct.

### Where it lives

`.github/workflows/ci.yml`, the `Mutation testing` job.

MalpenZibo:
**Throwaway. Do not merge. Opened only to make CI run the job in #253 on a diff that contains Rust, and will be closed as soon as it has.**

The pull request that actually fixes #253 touches only `.github/workflows/ci.yml` and `AGENTS.md`, and `cargo mutants --in-diff` on such a diff prints `Diff changes no Rust source files` and exits 0 without building anything. It would go green having exercised nothing — the very failure the change is about.

So this branch is #252's two commits (a real 245-line Rust diff, and the exact one whose mutation job died on `No space left on device` after 27 minutes) with the fix cherry-picked on top. Its `Mutation testing` job is the evidence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,8 +151,25 @@ jobs:
         env:
           BASE: ${{ github.event.pull_request.base.sha }}
         run: git diff "$BASE"...HEAD > /tmp/pull-request.diff
+      # `--lib --tests` because a plain `cargo test` also builds all sixty
+      # examples, and each of the `-j N` build directories compiles the crate
+      # from scratch — nothing is shared, and `--copy-target` is off by default,
+      # so there is no incremental reuse to soften it. Four trees at 26 GB on a
+      # runner with a few tens of gigabytes. Nothing is lost — the examples are
+      # compiled by the Test job, once, and they carry no tests of their own.
+      # No debug info for the same reason
+      # and because it is most of what remains: what matters here is whether a
+      # test failed, not where. Measured on a clean tree: 26 GB and 84s becomes
+      # 1.4 GB and 47s.
+      #
+      # It costs the doc tests, which target selection excludes and which cargo
+      # refuses to combine with `--doc`. A mutant only a doc test would have
+      # killed now reports as missed.
       - name: Mutate what changed
-        run: cargo mutants --in-diff /tmp/pull-request.diff -j 4 --no-shuffle
+        env:
+          CARGO_PROFILE_DEV_DEBUG: 0
+          CARGO_PROFILE_TEST_DEBUG: 0
+        run: cargo mutants --in-diff /tmp/pull-request.diff -C --lib -C --tests -j 4 --no-shuffle
       - name: Upload what survived
         if: failure()
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


MalpenZibo:
Evidence captured: `16 mutants tested in 9m: 3 missed, 13 caught` — a complete run where the same job on #252 died after 27m with one mutant of sixteen tested. Run: https://github.com/MalpenZibo/guido/actions/runs/33001169670/job/98283093304

Closing as planned. The Rust commits here belong to #252; the workflow fix goes up on its own.

Issue #268, opened 2026-08-27
A blur only the compositor performs is dropped for having no radius of its own


MalpenZibo:
## Observed

A container that asks only the compositor to blur its backdrop never gets a
blur region published, so the compositor blurs the whole surface or nothing —
never the container's shape.

`src/widgets/container/mod.rs:1729`:

```rust
if let Some(blur) = backdrop_blur
    && blur.radius > 0.0
{
    ctx.draw_backdrop_blur(..)
}
```

`radius` is the radius of the blur **guido** runs in its own shader. The
documentation on the flag that selects the other mechanism says it does not
apply — `src/backdrop.rs:43-45`:

> `COMPOSITOR` — What the compositor composites behind this surface, via
> `ext-background-effect-v1`. **The protocol carries no radius — the compositor
> picks its own — so `radius` does not apply here.**

So the code makes a value the documentation calls inapplicable into the switch
that decides whether the request exists at all. `BackdropBlur::new(0.0)
.sources(BackdropSources::COMPOSITOR)` — "do not blur anything yourself, let
the compositor blur what is behind me" — is exactly the spelling the docs
describe, and it emits no command.

Nothing downstream can recover from that. No `DrawCommand::BackdropBlur` means
`LayeredCommands::push` (`src/renderer/flatten.rs:373`) never sets
`compositor_blur`, so `paint_and_present`'s guard never calls
`sync_blur_region`, so `blur::regions_from_commands` — which does correctly
tessellate the rounded shape — is never reached.

### How it was confirmed

`examples/blur_example.rs` asks for exactly this, and says so in its own
comment: *"Restricted to the compositor's backdrop, since this example is about
the desktop showing through a translucent surface, not about blurring the
surface's own content."*

Run under `WAYLAND_DEBUG=1` on niri 26.04, changing only the radius:

| `BackdropBlur::new(..)` | `get_background_effect` | `wl_region` |
| --- | --- | --- |
| `0.0` | 0 | 0 |
| `1.0` | 1 | 52 |

At `0.0` the client binds `ext_background_effect_manager_v1`, receives
`capabilities(1)`, and then says nothing for the rest of its life.

What is on screen instead is the compositor's own rule — a niri
`layer-rule { match namespace="^guido-"; background-effect { blur true } }`
blurring the entire surface, which is what such a rule does when the client
never narrows it. On a compositor with no such rule the same example blurs
nothing at all, which is the same defect wearing the other face.

## Expected

[...]

Issue #264, opened 2026-08-27
A frame is plain data, but only a live compositor can open one


MalpenZibo:
The harness covers both ends of a frame and neither of the middle.

**Below it**, `tests/render_snapshots.rs:48` builds a `Tree` directly and calls
widget methods on it. **Above it**, `tests/golden_images.rs:189` renders to an
offscreen texture with the real `Renderer` — headless rendering already works,
and already runs in CI on lavapipe.

**Between them** is `App::run` (`src/lib.rs:1392`): three hundred lines that
weld the calloop event loop, the Wayland connection, the wgpu swapchain and the
job pipeline into one function. `create_wayland_app`
(`src/platform/wayland.rs:270`) returns `Err` without a compositor, and the
configure wait at `src/lib.rs:1472` blocks on `event_queue.blocking_dispatch`.
Nothing can construct an `App`, step it one iteration, or observe what it did.

What that leaves unwatched is not "Wayland". It is everything the library does
*with* what Wayland says:

- `dispatch_events` (`src/lib.rs:634`) routing an event through a *surface* —
  which widget is focused, what captured the pointer, and the surface state the
  event arrives with. Not widget-level hit testing: that is already reachable,
  and the correction below says so.
- the frame pacing gate — `frame_callback_pending` at `src/lib.rs:1253`
- the job pipeline: `distribute_jobs`, the orphan lane, `flush_bg_writes`,
  deadline scheduling
- surface configure, resize, scale change
- outputs appearing and disappearing. #209 parks this for want of a second
  monitor, but the guido-side half is a synthesized event, not a second monitor.
- **what the surface asks the compositor for** — exclusive zone, layer, anchor,
  margin, input region, blur region, popup reposition

That last group deserves separating, because it is the part a headless harness
can prove *completely*. Guido cannot prove niri interprets an exclusive zone
correctly — #203 is an assertion about niri, and no test in this repo can settle
it. Guido can prove that guido asked for 32.

## What a verification pass changed

Read at the sites on 2026-08-27, after this issue was first written from a
survey. Three things did not hold:

- **`#260` and `#261` are not examples of what nothing can reach.** They were
  found by a person clicking, but `tests/scrollbar_handle_tracking.rs` already
  dispatches events into a `Tree` with no compositor and asserts on handle
  geometry — including `a_track_click_moves_the_handle` and
  `pressing_the_painted_handle_starts_a_drag_rather_than_jumping_the_track`.
[...]

PR #269, opened 2026-08-27
The frame stops carrying a connection, and the phases stop passing one

MalpenZibo:
## What changed

**Step 1 of five in #264 — this does not close it.** The issue is implemented
one step at a time; the current step is the first one not done.

Wayland protocol types come out of the per-frame code path. `Frame` was a plain
description of what the compositor said — events, scale, size, two flags — that
also carried a live `WlSurface`, and `FrameContext` carried a
`&QueueHandle<WaylandState>` because two calls somewhere below wanted one. So
`layout_pass` and `paint_and_present` each opened by unpacking a Wayland type
they had no use of their own for, and `resolve_geometry` sat between them
touching neither.

**The design content the issue asked to be named first:** what replaces
`wl_surface` as the thing `paint_and_present` commits and requests a frame
callback on. The answer is that `WaylandState` owns its own `QueueHandle` — it
had none, which is why the handle was being threaded down from `App::run` — and
the frame path asks it by `SurfaceId` through three methods:
`request_frame_callback`, `damage_surface`, `mark_frame_callback_pending`. That
is `fn(&mut self, id: SurfaceId)`, the same spelling as the fifteen sibling
methods either side of them, and those three signatures are what step 2 turns
into the trait's first methods.

Two effects worth naming. `lib.rs` no longer contains the word `WlSurface`
anywhere; the import went with it. And `open_frame` used to clone that surface
handle for every frame it opened, *before* the pacing gate decided whether the
frame would be drawn — so a bar sitting still paid a clone per compositor
callback to throw it away. Paced-out frames no longer pay it, and `Frame` goes
from **104 bytes to 40** (measured, not estimated) for the ones that do.

## What proves it

`src/lib.rs`'s new `#[cfg(test)] mod a_frame_is_a_description_not_a_connection`
— four tests. It is #264's stated acceptance criterion for this step, and it
failed before the change with exactly the stated reason:
`error[E0063]: missing field 'wl_surface' in initializer of 'Frame'`.

Constructing a `Frame` is most of it. The rest is `paced_out`, the gate's four
conditions given a name — the one decision in the frame path that follows from
a `Frame` and a `Geometry` alone, and so the first one a test can reach now
that neither holds a connection. The reviewer mutated it six ways (each of the
four operands dropped, plus forcing the whole function `true` and `false`) and
all six were caught.

**Being precise about the criterion:** it reads "constructs a `Frame` *and
drives the frame path with it*". What landed constructs a `Frame` and drives
the pacing gate. `open_frame`, `resolve_geometry`, `layout_pass` and
`paint_and_present` are still unreachable from a test, because `FrameContext`
still needs a `ManagedSurface`, a `WaylandState` and a `Renderer`. That is the
honest maximum at step 1; steps 3 and 5 are where the rest becomes reachable.

Three commits, each compiling and clippy-clean on its own — the reviewer
verified this by extracting each commit's tree and building it, rather than
taking my word.

Harness: 711 tests, 0 failures; 9 goldens on lavapipe (passed, not skipped);
`cargo fmt --check`; clippy with `-D warnings`. No golden or snapshot moved,
and none should have — this draws no pixels differently.

## What the review found

`/simplify` ran before the commits, so its edits are inside them. Two findings
that mattered: **two `self.qh.clone()` calls I had written to dodge a borrow
conflict that does not exist** — `surfaces.get()` takes a shared borrow, and
edition 2024's precise closure captures make `&self.qh` fine inside
`get_or_insert_with` — and **four copies of one explanation**, three of them
added by me. Both fixed. Its reuse and altitude passes came back clean.

I **rejected** one finding: that the three new methods should call
`get_surface`/`get_surface_mut` rather than indexing `self.surfaces`. Two
agents disagreed about it, and the reviewer settled it — direct indexing is
used at six sites in `wayland.rs`, five in `popups.rs`, ten in `input.rs`; the
accessors are used only from outside `platform::wayland`. Clean in-module /
out-of-module split, and these are in-module.

The `reviewer` subagent then read the three commits. **One blocking finding:**
nothing proved the behavioural half. It replaced all three new methods with
no-ops — no frame callback ever requested, no damage ever reported — and the
entire suite passed, the four new tests included. That is accurate and it is
the point of #264: this layer has no sensor. The block was on the hand-run
being done and recorded, which is the next section.

Three findings worth answering, all acted on: naming the replacement in the PR
rather than only in a commit body (above); being explicit about what "drives
the frame path" came to mean (above); and not closing #264 (the body says so
and no commit uses a closing keyword). One note acted on: it measured `Frame`
at 104 → 40 bytes where a commit body of mine said "some forty bytes smaller",
estimating `size_of::<WlSurface>()`. Corrected on rebase.

Its notes on scope holding, and on the docs being right to leave alone
(`docs/ARCHITECTURE.md:146,160` describes protocol traffic, which is
byte-identical), are recorded here rather than acted on.

## What was checked by hand

niri 26.04, four examples.

- **`animation_example`** — springs run to their overshoot and settle, smoothly.
  This is the sensor that matters: an animation only advances if the loop keeps
  waking, and the loop wakes on the frame callback. A broken
  `request_frame_callback` would show as one frame drawn and everything frozen,
  with no error and green CI.
- **`text_input_example`** — typed and deleted; text appeared and disappeared
  cleanly, no stale glyphs underneath, nothing truncated. That is
  `damage_surface` on the `DamageRegion::Partial` branch, scale multiplication
  included.
- **`popup_example`** — positioning correct. `reposition_popup_if_changed` lost
  a parameter it never read.
- **`blur_example`** — **not a usable sensor on this machine**, and finding out
  why produced #268 below.

I first proposed `status_bar` here and was wrong: it is 36 lines of static text
and proves only that a surface configures and draws once. It cannot see the
frame callback re-arming, which is the thing most worth seeing.

## Left undone

**#268** — a blur only the compositor performs is dropped for having no radius
of its own. `container/mod.rs:1729` gates the whole `BackdropBlur` command on
`radius > 0.0`, where `radius` is guido's own shader radius and the
documentation on `BackdropSources::COMPOSITOR` says it does not apply to that
source. So `BackdropBlur::new(0.0).sources(COMPOSITOR)` — the documented
spelling for "let the compositor blur behind me" — publishes no region at all.
Confirmed with `WAYLAND_DEBUG=1`: at radius `0.0`, zero `get_background_effect`
and zero `wl_region`; at `1.0`, one and fifty-two. **Pre-existing** — verified
identical on `main` at 8f2f059 before attributing.

**Steps 2 through 5 of #264** stay in the issue, which carries an acceptance
criterion for each. Nothing this step defers was touched: `configured_size`,
`send_size` and `reconfigure` are untouched, no trait exists yet, the loop body
is still inline in `App::run`, and `src/lib.rs:1257` still reads
`Instant::now()` for the frame instant.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -737,6 +737,23 @@ struct Frame {
     force_render_surface: bool,
 }
 
+/// Whether this frame would only queue behind one the compositor has not shown.
+///
+/// The three that open the gate anyway — the fields say why each is what it is
+/// — are a surface with nothing on screen yet, a resize, and a scale change:
+/// all three have something to tell the compositor that cannot wait for a
+/// callback.
+///
+/// A function rather than the condition inline, because it is the one decision
+/// in the frame path that can be taken from a `Frame` and a `Geometry` alone,
+/// and so the one a test could reach first.
+fn paced_out(frame: &Frame, geometry: &Geometry) -> bool {
+    frame.frame_callback_pending
+        && !frame.force_render_surface
+        && !geometry.needs_resize
+        && !geometry.scale_changed
+}
+
 /// The physical size, and what about it changed since the last frame.
 struct Geometry {
     physical_width: u32,
@@ -1232,11 +1249,7 @@ fn render_surface(
     // loop iteration would re-ping the wakeup each time and spin the loop
     // flat out between frame callbacks. Input was dispatched above, so it is
     // never delayed by the gate; initialisation and resizes bypass it.
-    if frame.frame_callback_pending
-        && !frame.force_render_surface
-        && !geometry.needs_resize
-        && !geometry.scale_changed
-    {
+    if paced_out(&frame, &geometry) {
         render_stats::record_frame_skipped();
         return;
     }
@@ -2038,5 +2051,59 @@ mod dispatch_declares_the_moment {
             "the moment has to be cleared when the dispatch ends, got {after:?} \
              for an event from an hour ago"
         );
+    }
+}
+
+/// A frame is a description of what the compositor said, and the pacing gate is
+/// a decision taken from that description. Neither needs a compositor to exist,
+/// and until the Wayland types came out of the frame path neither could be
+/// named here: `Frame` held a live `WlSurface`, and a test has no display to
+/// get one from.
+#[cfg(test)]
+mod a_frame_is_a_description_not_a_connection {
+    use super::*;
+
+    fn frame(callback_pending: bool, force: bool) -> Frame {
+        Frame {
+            events: Vec::new(),
+            scale_factor: 1.0,
+            width: 200,
+            height: 50,
+            frame_callback_pending: callback_pending,
+            force_render_surface: force,
+        }
+    }
+
+    fn geometry(needs_resize: bool, scale_changed: bool) -> Geometry {
+        Geometry {
+            physical_width: 200,
+            physical_height: 50,
+            needs_resize,
+            scale_changed,
+        }
+    }
+
+    #[test]
+    fn a_frame_can_be_described_without_a_display() {
+        let f = frame(false, false);
+        assert_eq!((f.width, f.height), (200, 50));
+    }
+
+    #[test]
+    fn a_callback_still_in_flight_paces_a_frame_out() {
+        assert!(paced_out(&frame(true, false), &geometry(false, false)));
+    }
+
+    /// Each on its own, because any one of them alone has to open the gate.
+    #[test]
+    fn a_first_frame_a_resize_and_a_scale_change_each_bypass_the_gate() {
+        assert!(!paced_out(&frame(true, true), &geometry(false, false)));
+        assert!(!paced_out(&frame(true, false), &geometry(true, false)));
+        assert!(!paced_out(&frame(true, false), &geometry(false, true)));
+    }
+
+    #[test]
+    fn no_callback_in_flight_paces_nothing_out() {
+        assert!(!paced_out(&frame(false, false), &geometry(false, false)));
     }
 }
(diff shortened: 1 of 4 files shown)


MalpenZibo:
Mutation job: **35 mutants, 9 missed, 8 caught, 18 unviable.** Red, and reporting rather than blocking. Read rather than waved past, because the shape of the nine is the argument for #264 rather than a regression in this change.

The eight caught are all `paced_out` — the new test doing its job, and the same six mutations the reviewer ran by hand plus two more.

The nine missed are one list:

| | |
| --- | --- |
| `src/lib.rs:769` | `open_frame` → `None` |
| `src/lib.rs:788` | `delete !` in `open_frame` |
| `src/lib.rs:854` | `layout_pass` → `()` |
| `src/lib.rs:1069` | `paint_and_present` → `()` |
| `src/lib.rs:1238` | `render_surface` → `()` |
| `src/platform/backdrop.rs:95` | `sync_blur_region` → `()` |
| `src/platform/wayland.rs:372` | `request_frame_callback` → `()` |
| `src/platform/wayland.rs:386` | `damage_surface` → `()` |
| `src/platform/wayland.rs:394` | `mark_frame_callback_pending` → `()` |

Every one of them is "delete a function in the frame path and nothing objects". That is the same finding the `reviewer` produced by hand — it replaced the three new methods with no-ops and the whole suite stayed green — now measured rather than argued.

Three of the nine are new lines this change wrote, and six are pre-existing lines the diff dragged into the job’s scope. None of the nine is reachable by a test today: reaching any of them needs a `FrameContext`, which needs a `ManagedSurface`, a `WaylandState` and a `Renderer`. That is the hole, it is what #264 step 5 closes, and it is why this step’s hand-run on niri is in the body above rather than a formality.

Not filing an issue for it: #264 already is the issue, and its acceptance criterion is what turns these red.

Issue #270, opened 2026-08-27
A reservation that would evict every window on a monitor is deferred by a guard nothing watches


MalpenZibo:
## Observed

`publish_exclusive_zone` opens with a guard that defers publishing a reservation
until a content axis has been measured — `src/lib.rs:317-322`:

```rust
if config.exclusive_zone == surface::ExclusiveZone::Auto
    && measured.is_none()
    && surface::needs_content_measure(config)
{
    return;
}
```

Delete it whole and **714 tests pass** — the full `cargo test --all-features`,
lib and integration alike. Verified with a forced recompile, not an incremental
one.

Its own comment says what it is for, and the failure it describes is not a small
one:

> a surface declared `width(content()).anchor(TOP | LEFT | RIGHT)` is stretched,
> so 1920 is what the compositor imposed on an axis that was not ours.
> Re-anchor it into a side dock and that number becomes a reservation for the
> whole screen, pushing every other window off it, until the measure lands.

So the case the guard prevents is a surface that reserves an entire monitor and
evicts every window on it. Nothing would notice if the guard went.

The comment goes on to explain why only `Auto` defers — `Fixed`, `None` and
`Ignore` are constants, and deferring one defers it for ever, because nothing
republishes a policy that is not `Auto`. That second half is unwatched too.

## Expected

Removing any of the guard's three conditions fails a test.

## Acceptance

`src/lib.rs`'s `#[cfg(test)] mod a_second_host_can_answer_for_a_compositor`
already has the machinery: a `Recorder` implementing `SurfaceHost` that keeps
what `set_surface_exclusive_zone` was asked for. Three cases beside the three
there:

1. `width(content()).anchor(TOP | LEFT).exclusive_zone(Auto)` — a side dock
   whose reservation follows an unmeasured content axis — with `measured: None`
   publishes **nothing**. Deleting `needs_content_measure` from the guard
   publishes `SurfaceExtent::initial()`'s 1px, and deleting the whole guard
   publishes whatever the confirmed size happens to be. Both are wrong and
   neither fails today.
2. The same surface with `measured: Some(..)` publishes the measured width, so
   the deferral ends when the measure lands rather than for ever.
3. `exclusive_zone(48u32)` on the same content surface publishes 48 immediately.
   That is the `measured.is_none()` and `Auto` conditions together: a constant
   policy must not be deferred, because nothing would ever republish it.

Each case dies to one of the guard's three conditions, which is what makes them
three cases rather than one.

## Non-goals

[...]

PR #271, opened 2026-08-27
The frame path asks a trait, and Wayland is one answer to it

MalpenZibo:
## What changed

**Step 2 of five in #264 — this does not close it.**

Every request a frame makes of a compositor was a call to a concrete type, so
the shape of the asking was spread across five functions and readable only by
reading all five. `SurfaceHost` is that asking written down: fourteen requests
in four groups — what a frame is opened from, what layout asks for, the blur
region, and what present has to send. `WaylandState` implements it; `FrameContext`
and the frame functions are generic over it.

**One required method, thirteen defaulted.** `open_frame` has no default because
a host that cannot say what a frame is has nothing to host; the rest default to
doing nothing and knowing nothing, which is the honest answer for a host without
that protocol. What defaults cannot catch is a method added to the trait and
forgotten in `WaylandState` — it would silently no-op against a real compositor
rather than fail to build. The trait's doc says so.

**This is not the two-level shape agreed on the issue**, and that was raised with
the maintainer before the commits rather than presented after: the agreement was
a thin trait over a per-surface one, and this is flat, with a `SurfaceId` on
almost every method. The maintainer chose to keep it flat and split at step 3,
when a driver can say which shape survives use. Recorded as a comment on #264
with the two visible seams for whoever splits it — `supports_blur_region()`
takes no id because it is a fact about the process, and `batch_layer_requests`'
closure takes the whole host where a per-surface handle would take one surface.
Checked while deciding: the calls in `paint_and_present` are isolated from one
another, so a handle would be taken and dropped around each; the obstacle is
design time, not the borrow checker.

## What proves it

`src/lib.rs`'s `#[cfg(test)] mod a_second_host_can_answer_for_a_compositor` —
a `Recorder` implementing the trait, which is step 2's acceptance criterion, and
three tests. It failed before the change with the stated reason: `cannot find
trait 'SurfaceHost' in this scope`.

The tests assert something nothing anywhere has asserted before: **what a surface
asks the compositor for**, as opposed to what it draws. A bar reserving
automatically asks for the height it declared; a margin on the anchored edge is
part of that reservation; a policy that is not `Auto` sends nothing per frame.
Inverting the `Auto` guard fails all three; ignoring the margin fails one.

Two commits, each clippy-clean alone. It was going to be three — trait, callers,
recorder — but the trait without its callers fails `-D warnings` as dead code,
so trait and callers are one commit. The reviewer reconstructed that
intermediate rather than taking my word and got `error: trait 'SurfaceHost' is
never used`; my own three-way attempt reported `multiple methods are never used`
because its intermediate still had one caller. Either way the split is wrong.

Harness: 714 tests, 0 failures; 9 goldens on lavapipe; `cargo fmt --check`;
clippy with `-D warnings`. No golden or snapshot moved, and none should have.

Static dispatch throughout — no `dyn` anywhere in the diff, and the shipped
binary instantiates the generic frame path once, for `WaylandState`, exactly as
before. `cargo build --lib --all-features` after touching `src/lib.rs`: 1.34s
against 1.33s, three runs each, which is noise.

## What the review found

`/simplify` ran before the commits. Three findings acted on: **the trait had zero
defaults**, against the proportion #264 had already settled — fixed, and
`Recorder` fell from ~50 lines to 20; **needless `platform::WaylandState::foo(self, ..)`
qualification** on twelve methods, saying there was a name collision where an
inherent method already wins; and **an orphaned doc comment**, where deleting the
free `configured_size` left its rationale glued to `send_size`'s doc.

I rejected its claim that the six generic functions are unverified
generalisation — that is what step 2 is for, in the issue's own words. The
reviewer agreed and sharpened it: the generic callers are not an extra, they are
what stops the trait being dead code. Its caveat, worth stating once: only
`resync_exclusive_zone` and `publish_exclusive_zone` get a second instantiation
today, so "a wrong trait shows up here, where it is cheap" is half-collected in
this step and the rest arrives at 3 and 5.

The `reviewer` subagent then read the two commits. **Zero blocking findings.**
Three worth answering, all acted on:

1. **The same orphaned-doc defect, reintroduced by my own fix.** Inserting the
   trait between `Geometry`'s doc comment and `struct Geometry` left the trait
   documented as "The physical size, and what about it changed since the last
   frame" and `Geometry` with no doc at all. That is the third time in this
   session I have moved code and left its explanation behind.
2. **A false claim in a commit body.** It said the tests read a confirmed size.
   They do not — I verified that deleting `configured_size`'s only call from
   `publish_exclusive_zone` leaves the whole suite green. The reviewer proposed a
   `height(content())` case to fix it; I checked instead of applying, and it does
   not work: whenever the zone follows a `Content` axis the content-measure
   deferral returns first, so no case can reach `configured_size`. The commit
   body now says that plainly rather than claiming coverage.
3. **The manual run**, below.

Notes acted on: the module name overstated what it reaches, and a test name said
"the height it has" where the surface had only declared one — both renamed. And
it corrected my reasoning for keeping the trait in `lib.rs`: I said `Frame` being
lib.rs-private constrained it, which is false — a private item at the crate root
is visible in every descendant module, and it verified that. The real reason is
that `SurfaceHost` is the frame path's contract and the frame path is `lib.rs`.
It also disagreed with the agent who cited `docs/ARCHITECTURE.md:198` as
requiring the move; that passage is about `delegate_*` handlers.

## What was checked by hand

niri 26.04. This step re-routes fourteen protocol calls through a trait, and
nothing automated stands under the forwarding bodies.

- **`animation_example`** — springs reach their overshoot and settle. This is
  `request_frame_callback` arriving through the trait, not merely compiling: an
  animation only advances if the loop keeps waking on the callback.
- **`text_input_example`** — typed and deleted, clean, caret blinking. That is
  `damage_surface` on all three `DamageRegion` branches.
- **`popup_example`** — position and size correct: the only example that
  exercises `popup_auto_width` and `reposition_popup_if_changed`.
- **`blur_example`** — checked by protocol trace rather than by eye, because
  #268 means the example publishes nothing at radius zero. With the radius
  temporarily at `1.0`: one `get_background_effect` and 52 `wl_region`, the same
  numbers as before the trait. `sync_blur_region` forwards.

## Left undone

**#270** — `publish_exclusive_zone`'s content-measure deferral guard is watched
by nothing. Deleting it whole leaves all 714 tests green, verified by hand with a
forced recompile. Its own comment says what it prevents: a `width(content())`
surface re-anchored into a side dock reserving an entire monitor and evicting
every window on it. Pre-existing; this change is only what made it reachable by a
test for the first time, and #270's acceptance uses the `Recorder` this PR adds.

**#268** stays open — the blur defect found while verifying step 1.

**Steps 3 through 5 of #264**, each with its own acceptance criterion. Two things
step 3 inherits, both recorded on the issue: `pub(crate)` will not survive it,
because `App::run` is `pub` and a public item generic over a `pub(crate)` trait
fails `private_bounds` under `-D warnings`; and `configured_size` is on the trait
for `send_size`, which belongs to the surface-command path, because both paths
share `publish_exclusive_zone` and a generic caller cannot reach an inherent
method.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -269,10 +269,10 @@ fn close_surface_now(
 /// and log a line saying so, on every resize and every margin change.
 /// `measured` is for the one caller that knows a size the compositor has not
 /// been told about yet: the content-measure pass, which has just computed it.
-fn resync_exclusive_zone(
+fn resync_exclusive_zone<P: SurfaceHost>(
     id: SurfaceId,
     config: &SurfaceConfig,
-    wayland_state: &mut platform::WaylandState,
+    wayland_state: &mut P,
     measured: Option<(u32, u32)>,
 ) {
     if config.exclusive_zone == surface::ExclusiveZone::Auto {
@@ -295,10 +295,10 @@ fn resync_exclusive_zone(
 /// `Auto` reservation follows is always one we own — `follow_axis` gives up on
 /// an axis anchored to both edges, the only kind the compositor sizes — so the
 /// value we asked for is the value that will take effect.
-fn publish_exclusive_zone(
+fn publish_exclusive_zone<P: SurfaceHost>(
     id: SurfaceId,
     config: &SurfaceConfig,
-    wayland_state: &mut platform::WaylandState,
+    wayland_state: &mut P,
     measured: Option<(u32, u32)>,
 ) {
     // A content axis waiting to be measured has no size to reserve for, and the
@@ -320,7 +320,7 @@ fn publish_exclusive_zone(
         return;
     }
 
-    let known = measured.or_else(|| configured_size(wayland_state, id));
+    let known = measured.or_else(|| wayland_state.configured_size(id));
     let width = surface::requested_extent(config.width, known.map(|(w, _)| w));
     let height = surface::requested_extent(config.height, known.map(|(_, h)| h));
     let zone = config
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Mutation job: **53 mutants, 48 missed, 1 caught, 4 unviable.** Reporting, not blocking — but the number went from 9 missed to 48 between step 1 and step 2, and that deserves saying out loud rather than filing under "the job is advisory".

**26 of the 48 are new, and they are this change's own doing.** They all live in `impl SurfaceHost for platform::WaylandState`, and every one is the same question: replace a one-line forward with `()`, or a getter with a constant, and nothing objects. `request_frame_callback` → `()`. `damage_surface` → `()`. `sync_blur_region` → `()`. `supports_blur_region` → `true` *and* → `false`, both missed, which is as blunt as it gets.

That is the honest cost of introducing the trait: fourteen forwarding bodies where there used to be direct calls, and no test drives any of them. Only two things can — a real compositor, which is why this PR's hand-run traced the protocol rather than trusting the eye, and step 5's recorder-driven frame, which does not exist yet.

**The other 22 are the frame path, unchanged since step 1** reported the same shape: `open_frame` → `None`, `layout_pass` → `()`, `paint_and_present` → `()`, `render_surface` → `()`. They appear again because this diff touches more lines in those functions, so `--in-diff` mutates more of them.

**The 1 caught is `paced_out`**, from step 1 — the only line of this path a test can currently reach.

Nothing in the 48 points at a defect. `configured_size` → `Some((0, 1))` surviving is already documented in the second commit's body and is a property of the code, not a gap in the tests: no case can reach it, because whenever the exclusive zone follows a `Content` axis the content-measure deferral returns first. The rest are absence-of-tests at a layer that has none by construction.

Merging with this red on the record rather than quietly: the ratio gets worse before it gets better, and step 5 is what turns it around. If it does not, that is the argument for reconsidering the whole plan, and this comment is where the number to compare against lives.

Issue #256, opened 2026-08-27
An event is timed by whoever handles it, with whatever clock is nearest


MalpenZibo:
**The architecture is agreed before it is written.** This issue states the
problem and the alternatives; the shape is the maintainer's decision.

### Observed

A frame now declares its instant once and every animation in it is asked about
that moment (#250). Events do not: each handler that needs to know when
something happened reads the clock where it stands.

| where | what it times |
| --- | --- |
| `interaction.rs` :156, :223, :262, :316 | a ripple begins and ends with a press |
| `scroll.rs` :367, :437 | when a momentum began |
| `scrollable.rs:937` | the gap between two gesture samples, which *is* the velocity |
| `text_input.rs:96` | the window inside which two keystrokes are one undo |
| `text_input.rs:601` | the caret restarting on input |

Each is a real time-dependent behaviour and none of them can be driven from a
test. There is no way to say "press, wait 300ms, press again" and get the same
answer twice, so the coalescing window, the ripple's growth under a held press
and the velocity a flick is given are all observed only by sleeping and
asserting a band.

The momentum is the one that has already cost. It computes a speed from `dt_ms`
between samples (#248), and establishing that its tests were not flaky took
twenty consecutive runs on #243.

### Why this is not the same change as #250

**The time of an event is not an `Instant`.** Wayland sends it as a `uint` of
milliseconds — `wl_pointer.motion`, `.button` and `.axis` all carry
`"timestamp with millisecond granularity"`, and sctk surfaces it as
`PointerEventKind::Axis { time: u32, .. }`. Its epoch is not specified by the
protocol; it is the compositor's, not this process's monotonic clock. There is
no `duration_since` between that and an `Instant` without choosing an anchor,
and a `u32` of milliseconds wraps in about 49 days.

Nor can a local `Instant` simply stand in for it: events are pushed onto
`pending_events` and handled later, so *when it arrived* and *when it is read*
are two different moments. That gap is why the protocol carries a timestamp at
all.

And `Event` is public API (`src/widgets/widget.rs:501`). Giving its variants a
timestamp changes how every `match` on an event is spelled, inside the crate
and outside it — `tests/external_widget.rs` and the book included.

### What the protocol actually hands us, checked

Every device carries a timestamp, and one of them is not what it looks like.

| device | where the time is | what it is |
| --- | --- | --- |
[...]

Issue #250, opened 2026-08-26
An animation reads the clock for itself, so nothing outside its module can watch it move


MalpenZibo:
### Observed

`AnimationState::advance` takes no time. It reads the clock through two private
fields — `start_time`, set from `Instant::now()` at `animations.rs:101` and
`:163`, and `timeline.playing`, set at `:411` — and `advance_timeline` calls
`started.elapsed()` at `:427`.

The module's own tests never sleep. `at()` (`:820`) and `play_at()` (`:658`)
rewind those fields by a `Duration`, and the docstring above the first says
exactly why:

> `advance` reads the clock through `start_time`, so moving that back is the
> whole simulation. Sleeping instead would make the interruption point depend on
> how loaded the machine is — and the interesting half of a spring's phase space
> is on the far side of its overshoot, which a stretched sleep wanders into by
> accident.

That cure reaches exactly as far as the file. `characterization.rs` is a sibling
module and cannot touch a private field, so `settle` (`:1853`) sleeps 4 ms a
frame and can only observe where an animation *lands*.

**Measured.** With `TimingFunction::Linear` changed from `t` to `t * t` — the
same endpoints, the wrong curve — all 18 integration binaries stay green
(~120 tests), and so does every test in `characterization.rs`. Six unit tests
fail, and all six are in the three modules that can name an instant:
`timing.rs`, `keyframes.rs`, `animations.rs`. Nothing outside them can tell a
transition that eases from one that does not.

The cost has been paid twice this week. #243 ran four tests twenty consecutive
times to establish they were not flaky, and its review found an assertion there
about 30 ns wide, which held only because two `Instant::now()` calls in adjacent
statements sit further apart in a debug build than in a release one.

### Expected

Time arrives as an argument, the way `ripple.rs` already does it. Nothing in its
production code reads the clock — `advance`, `release`, `opacity_at`, `gone` and
the rest all take `now: Instant` (`ripple.rs:196-326`) — and
`Container::advance_animations` reads it once for them at `mod.rs:1362`.

The same read serves the declared animations, so one call to
`advance_animations` is one instant. Today it is several: each `AnimationState`
calls `elapsed()` for itself, and the ripple's `Instant::now()` is later still.

### Acceptance

**A1 — a mid-flight assertion from outside the module.** A new test in
`characterization.rs`: a container with `animate_width` on a 100 ms `Linear`
transition, driven from 0 to 100 and advanced to exactly 50 ms, asserts `50.0` —
[...]

PR #272, opened 2026-08-27
The waiting is calloop's, the working is guido's, and they are two functions

MalpenZibo:
## What changed

**Step 3 of five in #264 — this does not close it.**

`App::run`'s loop did two jobs: computed a timeout and blocked in
`event_loop.dispatch`, then ran a hundred and twenty lines of everything guido
does with what came back. A driver that wants to step the application forward
wants the second half and not the first, and there was no way to ask for it.

Three things landed:

- **`Pending` and `wait_for`** — the sleep decision, which was four conditions
  inline and reachable only by running a compositor.
- **The frame instant is handed to `render_surface`** rather than read inside
  it, sampled once for the pass. That is the residual #250 and #256 left behind.
- **`iterate`** — the working half, a free function taking a `LoopContext` by
  value. Free rather than a method because `render_surface`,
  `process_surface_commands` and `sync_platform_state` beside it all name the
  state they touch, and this is the one change whose point is naming what a
  driver must supply. It needed two things from `App`, so it takes those two.

## What proves it

`src/lib.rs`'s `#[cfg(test)] mod what_the_loop_waits_for` — five cases on
`wait_for`, red before the change with `cannot find function 'wait_for' in this
scope`. The reviewer ran `cargo mutants` on it: **4 mutants, 4 caught.**

They are the shape of the decision rather than a sample: idle blocks, each of
the three flags alone polls, a deadline is slept to exactly, one already gone is
not a negative sleep, and pending work outranks a later deadline. The subtle one
is `wake_requested` — it lands *after* the iteration consumed the flag, so
blocking on one already set would suppress every later ping. That was carried by
a comment and nothing else.

**The step's acceptance criterion is half met, and this is the half that is
missing.** It asks that "an in-crate test drives one iteration and observes that
it ran". Nothing drives `iterate`: `LoopContext` names a `Connection` and a
`QueueHandle`, and four of the functions it calls — `process_session_lock`,
`process_surface_commands`, `init_pending_gpu`, `sync_platform_state` — take
concrete Wayland types behind no trait. `cargo mutants` on `iterate`: **5
mutants, 5 missed.** Deleting the `!` on `process_surface_commands` — quit on
every successful pass, never quit when the commands say to — is invisible to 719
tests. That gap and its four call sites are written on #264 so step 5 meets it
knowingly rather than halfway through.

Two more things the tests do not watch, said because "tested five ways" is the
phrase a reader carries away: they watch `wait_for` and not its four inputs
(hardcoding `force_render: false` at the gathering site leaves all five green,
and gathering needs a `WaylandState`), and nothing proves the frame-instant
commit's two claims in-crate.

Harness: 719 tests, 0 failures; 9 goldens on lavapipe; fmt; clippy with
`-D warnings`. Three commits, each clippy-clean alone — the reviewer rebuilt
each rather than taking my word. No golden or snapshot moved.

## What the review found

`/simplify` ran before the commits, and three of its four findings were mine to
own: **a doc comment attached to `LoopContext` that described `iterate`** — the
fifth time this session I have moved code and left its explanation behind, which
is now written down where I will read it next time; **redundant `&mut *ctx.foo`
reborrows**, from not hoisting locals the way `layout_pass` and
`paint_and_present` beside it do; and **`ControlFlow` buying nothing over
`Option`**, every use being the two-arm degenerate case. Its altitude pass said
`iterate` should be a free function like its siblings; I agreed and changed it.

The `reviewer` subagent then read the three commits. **One blocking finding: the
hand-run, below.** Three worth answering:

1. **The extraction moved the frame instant's sample point**, from just before
   the render pass to the top of the iteration — ahead of the session lock, the
   surface commands, and on the iteration that builds one, the whole of
   `Renderer::new`. Bounded, because every consumer saturates, but the commit
   body read as a pure extraction and that line was not. Fixed rather than
   disclosed: `iterate` takes `Option<Instant>`, the loop passes `None`, and the
   render pass takes the moment where the frame is. A driver that names one is
   still obeyed. The fix went **into** the commit, not on top of it — a fourth
   commit correcting the third is the tidy-up commit the atomic-commits criterion
   exists to catch.
2. **A commit body claimed a measurement I had not made.** "Measured: an atomic
   load and a scan…" was reasoning, not a benchmark. It says "reasoned" now.
3. The `see wake_request_pending` cross-reference was lost when that comment
   became a doc; restored.

Its note that `LoopContext` could be taken by value and destructured — shorter
than the hoist `/simplify` asked for — is applied. Its note that `wait_for`
returns 16 ms even when a deadline is sooner, so that deadline slips by up to a
frame, is **pre-existing**: the `needs_polling` branch it replaced did the same.
Not changed here.

It also confirmed the extraction is behaviour-preserving: it read every moved
line against `origin/main` and found the phase order, the argument order, the
borrows and all four exit paths unchanged. And it checked specifically for the
orphaned-comment defect across every moved block — clean this time.

## What was checked by hand

niri 26.04, three examples, chosen because `iterate` has no automated sensor at
all and the reviewer mapped which example covers which moved call site.

- **`multi_surface`** — top bar and bottom dock, sharing one signal. Both
  appeared and stayed live, and the shared count updated on both. That is the
  closest thing available to the frame-instant commit's "two bars drawn together
  are the same moment".
- **`surface_runtime`** — anchor, size, margin and reservation changed at
  runtime: the branch behind the two surviving `!` mutants
  (`process_surface_commands`, `init_pending_gpu`, the lazy renderer).
- **`status_bar`** — the surface appears and stays, which is the lazy
  `Renderer::new` inside `iterate` and the steady loop after the split.

All three reported fine by the maintainer. I asked specifically whether
`surface_runtime`'s reservation tracked the bar without drifting and did not get
a more specific answer than "ok", so that detail rests on nothing visibly wrong
rather than on a deliberate look.

`simple_lock` would cover `process_session_lock`, the fourth moved call site,
and was **not run** — it locks the screen, which is why #209 skipped it too.
That call site is unverified by anything.

## Left undone

**Step 5's scope is larger than the issue says**, recorded as a comment on #264
with the four functions and their signatures. Making `iterate` generic over
`SurfaceHost` would not be enough: the trait covers the frame path, and these
four are the surface-command path this issue deliberately left out. Either they
move behind it too, or the harness drives something narrower than a whole
iteration. No new issue — #264 is the issue, and this is a scope note on a step
that has not started.

**`process_session_lock` is the one moved call site nothing checked**, by test or
by hand. Also on the issue comment.

**#268** and **#270** stay open, both found earlier in this series.

**Steps 4 and 5 of #264.** One flag for step 4: the lazy `Renderer::new` is
inline inside `iterate` and `LoopContext` holds `&mut Option<Renderer>` for it —
carried over unchanged, but one layer deeper than it was.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1461,9 +1461,10 @@ fn render_surface<P: SurfaceHost>(
     // one frame several instants, and made the middle of an animation something
     // no test could ask about — only sleep towards.
     //
-    // Handed in rather than read, and sampled once for the pass rather than
-    // once per surface: two bars drawn together are the same moment, and a
-    // driver that wants to say when a frame happened has somewhere to say it.
+    // Handed in rather than read, and sampled once per iteration rather than
+    // once per surface: two bars drawn in the same pass are the same moment,
+    // and a driver that wants to say when a frame happened has somewhere to
+    // say it.
     ctx.tree.set_frame_instant(Some(at));
 
     run_jobs(root, ctx.tree, layout_roots, active_roots);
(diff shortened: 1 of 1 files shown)


MalpenZibo:
Mutation job: **11 mutants, 4 missed, 4 caught, 3 unviable.** No surprises — every one of the four missed is already named in the body above.

| | |
| --- | --- |
| caught | all four are `wait_for` — CI agrees with the reviewer's local 4/4 |
| missed | `render_surface` → `()`, `iterate` → `None`, and the two `!` deletions inside `iterate` (`:1831`, `:1883`) |

The trend across this series, since step 2's comment asked for a number to compare against:

| step | missed / tested |
| --- | --- |
| 1 — Wayland types out of the frame path | 9 / 35 |
| 2 — the trait | **48 / 53** |
| 3 — the loop splits | 4 / 11 |

Step 2's spike was its own doing and was recorded as such: fourteen forwarding bodies added at once, none drivable by a test. It was not a slide that kept going. This step adds one function that is fully covered and moves a body that was already uncovered, so `--in-diff` sees mostly the former.

What has not moved is the floor. `iterate` → `None` — the whole iteration does nothing, no session lock, no surface commands, no render, no flush — survives 719 tests, and so does inverting the exit condition on `process_surface_commands`. That is the same hole steps 1 and 2 reported, now with a name and a signature, and step 5 is still what closes it. The examples run on niri are what stands in for it today, and `process_session_lock` is the one moved call site not even that reached.

Issue #273, opened 2026-08-27
The renderer skill describes one of the two places a frame can land


MalpenZibo:
`.claude/skills/renderer/SKILL.md:73-79` has a section called *Rendering somewhere other than a screen*. It says:

> `render()` is that, plus acquiring a swapchain texture and presenting.

That was true of one function. Since #264's step 4 it is true of one *branch*: `Renderer::render` takes a `RenderTarget`, and a frame lands either on a compositor's swapchain or on a texture the caller owns. The skill does not mention the second, or that it lives behind a `testing` cargo feature, or that `OffscreenTarget::read_pixel` is how a target is read back.

## Observed

The section describes the shape the renderer had before `RenderTarget` existed. Nothing it says is false — `render_to_view` is still what it was — but an agent reading it to find out how to draw without a screen is told to call `render_to_view` and build its own target, which is what `tests/golden_images.rs` does and what `src/lib.rs`'s in-crate test deliberately stopped doing.

## Expected

The skill names both targets, says which cargo feature the offscreen one needs, and points at `read_pixel` — so the next change that wants a frame without a compositor finds the path that already exists rather than building a third one.

## Acceptance

`tests/documentation_references.rs` passes today and will keep passing either way: it reads API names in backticks and no name became wrong. So this cannot be proved by the harness, and the check is that the section names `RenderTarget`, `OffscreenTarget` and the `testing` feature, read by whoever reviews the change.

Worth doing at the same time: #264's step 5 adds `tests/headless_app.rs`, which is the first consumer outside the crate and the first thing that has to turn the feature on. The skill should describe what that test does once it exists.

## Non-goals

- Not `docs/RENDERER.md` or the book: neither names any of this, and neither should — a testing-only target is not user documentation.
- Not the `testing` feature's design, which #264 settled.

## Where it lives

`.claude/skills/renderer/SKILL.md:73-79`.

---

Surfaced by the `reviewer` subagent on #264's step 4, and left undone there deliberately: the skill is what the next renderer change reads, and it deserves a considered pass rather than a line patched into a pull request about something else.

PR #274, opened 2026-08-27
A frame lands where its surface points, and one of the two is a texture

MalpenZibo:
## What changed

**Step 4 of five in #264 — this does not close it.**

`ManagedSurface` held a swapchain, so a frame had exactly one place it could go.
`Renderer` already knew better — `render_to_view` has drawn into a caller's
texture since the goldens needed one — but nothing above it could choose, so
everything from `paint_and_present` upward needed a compositor to run at all.

`RenderTarget` is the choice: a swapchain is acquired, drawn into and presented;
a texture is drawn into and kept. `render` takes the second in one line, because
`render_to_view` was already the first one's body with the acquire and the
present taken off.

**The `testing` cargo feature now exists**, off by default, which resolves half
of what step 2 left open. The texture half of `RenderTarget` and
`OffscreenTarget::read_pixel` live behind `#[cfg(any(test, feature = "testing"))]`
— a normal build should not acquire a render-to-texture API on the way to
letting a test read a pixel. In-crate tests reach it through `cfg(test)` alone;
step 5's test is outside the crate and will need the feature.

The target carries no size and no format of its own: `wgpu::Texture` knows both,
and a fact derived cannot drift from a fact stored.

## What proves it

`src/lib.rs`'s `#[cfg(test)] mod a_frame_lands_where_the_surface_points`, two
tests. **Step 4's acceptance criterion is met literally — the first time in this
series.** A pixel the application drew is read back without a compositor, and it
arrives through `render_surface`: opening the frame, resolving the geometry, the
pacing gate, layout, paint, flatten, damage and present.

Both verified by inversion. Changing the surface's background changes the byte
read. Replacing `RenderTarget::resize`'s offscreen arm with the no-op it first
was turns the second test red. Mutating the readback stride or its offset turns
the first red.

That last one only after a correction the reviewer's mutation run forced: the
first draft used a 100-pixel width written as 50, and at 50 pixels a row is 200
bytes, which pads to 256 whatever arithmetic produced it — so a broken stride
walked straight through an assertion whose comment claimed to be watching for
exactly that. At 100 the 400 bytes pad to 512 and the mutation dies.

Harness: 721 tests, 0 failures; 9 goldens on lavapipe; fmt; clippy with
`-D warnings`; and `cargo check` clean in `--no-default-features`, default,
`--features testing` and `--all-features` — the gating asymmetry is the failure
mode a new cargo feature has, and I introduced one (an ungated `pub use` of a
gated type) before finding it.

## What the review found

`/simplify` first, three findings, all mine: **a stale doc comment left stacked
above the new one** on `Renderer::render`, so the published doc claimed both
"nothing reached the screen" (untrue for a texture, which cannot fail that way)
and the new explanation — the sixth orphaned-comment defect this session;
**`OffscreenTarget` storing width and height** the texture already knows, which
is *why* `format()` had to write out `Rgba8Unorm` a third time; and **the test
hardcoding that format** where the accessor it was exercising would answer.

Its altitude pass argued `resize` declining for a texture is a trap rather than
a simplification, and it is right: `resolve_geometry` compares the target's size
to the frame's and calls `resize` when they differ, so a target that quietly
kept its own size would report a resize on every frame for ever, repaint the
whole tree each time, and defeat the incremental paint
`tests/paint_cache_across_frames.rs` exists to protect. It reallocates.

The `reviewer` subagent then read the two commits. **Two blocking findings, both
acted on:**

1. **The one decision in the change was executed by nothing.** `cargo mutants`
   on `gpu_context.rs`: `replace RenderTarget::resize with ()` — the exact
   spelling I had just rejected — survived the whole suite, along with every
   comparison in its guard. My doc comment was the only defence, and prose is
   not a sensor. Fixed with the second test, whose second assertion is the
   entire argument.
2. **The new test would never have run in CI.** The `test` job installs no
   Vulkan ICD, so `GpuContext::try_new()` returns `None`, the test prints
   "skipping" and passes; the `golden` job has lavapipe but runs
   `--test golden_images`, which excludes lib unit tests. Neither job would have
   executed a line of this change. Fixed: the golden job now also runs
   `cargo test --all-features --lib` with `GUIDO_GPU_REQUIRED=1`, so a skip
   there is a failure — the rule `GUIDO_GOLDEN_REQUIRED` already applies one
   line above.

Three worth answering. Its point that the test stepped over `open_frame`,
`resolve_geometry`, the gate and `layout_pass` is fixed by the same rewrite —
the test drives `render_surface` now. Its note that the frame reaching the
texture contains no widget output, so what is proved is the clear colour and not
that the paint pass arrived, is true and stated here. And it confirmed my
reasoning for declining a reuse suggestion: making `tests/golden_images.rs` call
`RenderTarget::offscreen` would tie an external test to the `testing` feature and
plain `cargo test` would stop compiling. It named two ways out for step 5 to
settle — a self dev-dependency, or `required-features` on the test target, the
second of which turns a missing feature into a silently skipped one.

## What was checked by hand

niri 26.04. This change alters the type every production surface holds and the
entry point every production frame goes through, and the swapchain arm has no
automated cover at all.

- **`status_bar`** — the surface appears and stays. If the `Swapchain` arm of
  the new `match` were wrong, nothing would be on screen.
- **`animation_example`** — springs reach their overshoot and settle: present
  running sixty times a second rather than once.
- **`surface_runtime`** — anchor, size and margin changed at runtime, which is
  what drives `RenderTarget::resize`'s swapchain arm, the line changed last.

All three reported fine by the maintainer.

## Left undone

**Nothing can create a headless surface through the real path**, recorded on
#264. `ManagedSurface::init_gpu` only ever builds `RenderTarget::Swapchain`, via
a call needing a real window handle; the test reaches an offscreen target by
overwriting the field after construction, bypassing `SurfaceManager`. So the
test goes through the frame path but not through initialisation, and how a
surface is *born* headless is step 5's, on top of what the issue assigned it.
That moves the boundary between 4 and 5 from where the issue drew it.

**`SurfaceHost` is still `pub(crate)` while `App::run` is `pub`** — step 2's
other inherited item. The `testing` feature is now the obvious answer, but
nothing has forced the decision.

**`.claude/skills/renderer/SKILL.md:73` is half stale** — its "Rendering
somewhere other than a screen" section says `render()` is `render_to_view` plus
acquire and present, true now only of the swapchain arm, and it does not mention
the offscreen target. Not edited here because the skill is what the next
renderer change reads and it deserves a considered pass rather than a line
patched in; **#273** files it.

**#268** and **#270** stay open. **Step 5** remains.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,15 @@ jobs:
         run: vulkaninfo --summary | head -n 40
       - name: Run golden image tests
         run: cargo test --all-features --test golden_images
+      # The unit tests that need an adapter live in the lib, and the `test` job
+      # has no ICD — so they skip there and report green. This is the job that
+      # has one. `--lib` runs the whole set, which is cheap, and
+      # `GUIDO_GPU_REQUIRED` makes a skip a failure here for the same reason
+      # `GUIDO_GOLDEN_REQUIRED` does above it.
+      - name: Run the unit tests that need a GPU
+        env:
+          GUIDO_GPU_REQUIRED: 1
+        run: cargo test --all-features --lib
       - name: Upload the pixels that moved
         if: failure()
         uses: actions/upload-artifact@v4
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,11 @@ license.workspace = true
 [features]
 default = ["svg", "webp"]
 render-stats = []
+# The apparatus #264 needs to drive a frame with no compositor: an offscreen
+# render target and the readback that makes one worth drawing into. On by
+# nothing, so a normal build ships no render-to-texture API it did not ask for;
+# in-crate tests get it from `cfg(test)` without the feature at all.
+testing = []
 # SVG rendering via resvg (pulls a second text-shaping stack + software
 # rasterizer). Without it, ImageSource::Svg* sources fail to decode with a
 # warning instead of failing to compile.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2494,3 +2494,169 @@ mod what_the_loop_waits_for {
         assert_eq!(wait_for(&pending, now), Some(FRAME));
     }
 }
+
+/// A frame drawn by the application, landing somewhere a test can read.
+///
+/// `tests/golden_images.rs` already renders without a compositor, but it builds
+/// its own target and calls the renderer itself — so it proves the shaders and
+/// nothing above them. This goes through `render_surface`: opening the frame,
+/// resolving the geometry, the pacing gate, layout, paint, flatten, damage and
+/// present. What it reads back is what the compositor would have been handed.
+#[cfg(test)]
+mod a_frame_lands_where_the_surface_points {
+    use super::*;
+    use crate::renderer::{GpuContext, RenderTarget};
+    use crate::surface::SurfaceConfig;
+    use crate::widgets::{Color, container};
+
+    const W: u32 = 100;
+    const H: u32 = 32;
+    const BG: Color = Color::rgb(0.25, 0.5, 0.75);
+
+    /// A host with one surface and nothing else. Every other request the frame
+    /// path makes is answered by the trait's defaults.
+    struct OneSurface {
+        width: u32,
+        height: u32,
+    }
+
+    impl SurfaceHost for OneSurface {
+        fn open_frame(&mut self, _id: SurfaceId) -> Option<Frame> {
+            Some(Frame {
+                events: Vec::new(),
+                scale_factor: 1.0,
+                width: self.width,
+                height: self.height,
+                frame_callback_pending: false,
+                force_render_surface: true,
+            })
+        }
+    }
+
+    /// A surface pointing at a texture, and the pieces the frame path wants.
+    fn surface_on_a_texture(
+        gpu: &GpuContext,
+        tree: &mut Tree,
+        texture: (u32, u32),
+    ) -> surface_manager::ManagedSurface {
+        let (widget, owner) = reactive::with_owner(|| Box::new(container()) as Box<dyn Widget>);
+        let mut surface = surface_manager::ManagedSurface::new(
+            SurfaceId::next(),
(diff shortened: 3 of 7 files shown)


MalpenZibo:
Mutation job: **48 mutants, 41 missed, 7 unviable — zero caught.** Zero, not few, and the number is an artefact rather than a measurement.

The job installs `libwayland-dev` and `libxkbcommon-dev` and no Vulkan ICD (`ci.yml`, the `mutants` job). Both tests this pull request adds call `GpuContext::try_new()`, get `None`, print "skipping" and pass — so every mutant they would kill survives by construction. `RenderTarget::resize` → `()` is in that list, and it is the one mutant I verified dies locally and which the second test exists to kill.

The same blindness is what the reviewer's second blocking finding was about, and the fix landed one job over: the `golden` job now runs `cargo test --all-features --lib` under lavapipe with `GUIDO_GPU_REQUIRED=1`, and its log shows both tests running rather than skipping:

```
test a_frame_lands_where_the_surface_points::a_target_the_wrong_size_is_resized_once_and_not_again ... ok
test a_frame_lands_where_the_surface_points::a_surface_background_reaches_the_texture_it_points_at ... ok
test result: ok. 583 passed; 0 failed
```

So the tests are watched. The job that watches the watchers is not, and it is now systematically wrong in one direction for anything the renderer touches: it will report every GPU-dependent mutant as missed for ever, which understates coverage precisely where a golden or a readback is the only sensor there is. Merging with that on the record rather than quietly, because the number in this comment is the one somebody will compare against later, and it deserves the asterisk.

Not fixed here: the job would need `mesa-vulkan-drivers` and the ICD dance the golden job already does, which is a change to the sensor-of-sensors rather than to this change. Worth its own issue if the ratio ever starts being used for anything.

For the series so far, with that caveat attached to the last row:

| step | missed / tested |
| --- | --- |
| 1 | 9 / 35 |
| 2 | 48 / 53 |
| 3 | 4 / 11 |
| 4 | 41 / 48, all of them for want of an adapter |

Issue #283, opened 2026-08-29
The headless tests have never met an adapter in CI, and skip green in every run


MalpenZibo:
## Observed

`tests/headless_app.rs` has never run against a GPU adapter in CI. Every one of
its six tests skips itself in every run, and reports green.

Two jobs, and the target falls between them:

| job | adapter | what it runs |
| --- | --- | --- |
| `test` | **none** | `cargo test --all-features` — builds and runs `headless_app`, which then skips |
| `golden` | lavapipe | `--test golden_images` and `--lib` only |

`Headless::new()` returns `None` without an adapter and each test returns
early. `GUIDO_GPU_REQUIRED` exists to make that a failure and is set in the
`golden` job — on the two steps that do not include this target.

So the harness #264 was built for, and every assertion #279 and #282 added to
it, is watched by nothing but a developer's own machine.

## Why it went unnoticed

The same shape twice. #277 found that the `test` job has no ICD, so the GPU unit
tests in `--lib` skipped there; the fix was to run `--lib` in the `golden` job
with `GUIDO_GPU_REQUIRED=1`. That closed the `--lib` half and left the
integration half open, because the golden job names its targets one at a time
and nobody added the third.

`cargo test --all-features` in the `test` job *does* run the file. It is the
adapter that is missing, not the invocation — which is why the job is green and
the file's own skip message scrolls past in a log nobody reads.

## Expected

The headless tests run against lavapipe in CI, and a skip there is a failure,
for the same reason it is a failure for the goldens: a test that needs an
adapter and passes without one reports green for the single reason it exists.

## Acceptance

- The `golden` job runs `tests/headless_app.rs` with `GUIDO_GPU_REQUIRED=1`.
- Confirmed by making one headless test fail and watching that job go red — not
  by reading the workflow. A step that runs the wrong target looks identical to
  one that runs the right target and finds nothing.
- The job's wall clock before and after is stated in the pull request. It is
  cheap now (six tests, 167ms with the default threads after #281) and the
  number is worth having before #275 grows the file.

## Non-goals

- Not giving the `test` job an adapter. One job with a rasterizer is the shape
  this repository chose, and the answer is to name the target there.
- Not the question of which other targets need the same treatment — but
  `tests/text_at_angles.rs` builds a `GpuContext` too and is worth a look while
  the workflow is open.

---

Found by the `reviewer` subagent on #281 and confirmed by reading the workflow:
`.github/workflows/ci.yml` lines 67, 112 and 121.

[...]

PR #287, opened 2026-08-29
TEMPORARY: proving the golden job runs the headless tests (do not merge)

MalpenZibo:
Throwaway. #283's acceptance asks for the fix to be confirmed by breaking a headless test and watching the job go red, not by reading the workflow. One assertion in `tests/headless_app.rs` is deliberately wrong on top of the real change.

Expected: **Golden images** red, and `Test` green — because the `test` job has no adapter, which is the whole point.

Closed as soon as it has reported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,12 @@ jobs:
     name: Golden images
     runs-on: ubuntu-latest
     env:
-      # Without an adapter the golden tests skip themselves, which would report
-      # green here for the one reason that matters. In CI a skip is a failure.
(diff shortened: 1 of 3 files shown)


MalpenZibo:
Done its job: **Test** green, **Golden images** red on `a_bar_reserving_automatically_declares_its_height_when_it_is_created` — the assertion this branch deliberately broke. The same broken test passing in one job and failing in the other is #283 and its fix in one run.

For the record, what the rasterizer job ran: lib 590, `golden_images` 9, `headless_app` 5 passed 1 failed, `text_at_angles` 2 — all with real counts rather than skips.

Issue #316, opened 2026-09-01
A cull rect follows a child to where it was laid out, not to where its transform draws it


MalpenZibo:
## Observed

`paint_children` hands each child a cull rect translated into that child's
coordinate space by the child's laid-out origin, and by nothing else
(`src/widgets/paint_children.rs:166-172`):

```rust
if let Some(ref cull) = opts.cull_rect {
    child_ctx.set_cull_rect(Rect::new(
        cull.x - child_x,
        cull.y - child_y,
        cull.width,
        cull.height,
    ));
}
```

A child's own `translate`, `rotate` or `scale` is not composed into it. So a
child that draws somewhere other than where it was laid out is culled — and now
also windowed — against a rect describing where it *would* have been.

A wrapper carrying `.translate(Translate::new(0.0, -300.0))` inside a 200 px
vertical scroller settles at 7 rows painted, and they are the wrong seven: the
rows the rect names, not the rows the translate brought into view.

## What #204 changed about it, and what it did not

Found by the `reviewer` subagent on #204, which introduced the binary-search
window in `paint_children`. It is worth being exact about the blame, because
the numbers look like a regression and are not:

- **Steady state is identical, and was already wrong.** Probed both ways with a
  two-frame harness driving `cache_paint_results`, as
  `tests/paint_cache_across_frames.rs` runs it: after one scroll event, `main`
  and the #204 branch both settle at the same wrong 7 rows.
- **What moved is one frame.** On `main` the *first* frame painted 20 of 20,
  because the per-child cull only drops children the paint cache calls clean
  and on a first paint nothing is. The window drops them unconditionally, so
  the branch paints 8 on that frame too. `main` was accidentally right for
  exactly one frame and #204 spends it.

So this is older than #204 and #204 makes it visible one frame sooner. The
comment at `src/widgets/paint_children.rs:110` says the window keeps one child
either side because a transform moves what a child draws off its laid-out
bounds — true as a margin, and it does not bound a 300 px translate. The
comment now says so rather than implying otherwise, which is the honest half of
this that #204 did land.

## Expected

A child is culled, and windowed, against the rect its drawn geometry actually
occupies — so a transformed child that is on screen is painted, and one that is
off screen is not.

## Acceptance

A test in `tests/paint_cache_across_frames.rs`, which is where the cross-frame
behaviour lives, or beside the new `tests/scroll_virtualization.rs`: a
scroller 200 px tall over rows whose wrapper carries
[...]

PR #317, opened 2026-09-01
The visible window belongs to whoever has the rect, not to whoever scrolls

### LINKED ISSUE #204: A scrollable list loses its virtualization if the rows are wrapped: 750x more work, silently
`Container::paint` narrows a scrollable container's children to the visible
window with a binary search over their bounds, so a long list paints only what
is on screen. It only ever sees the container's **direct** children, so wrapping
the rows in a layout container turns it off completely, silently.

Both scrollable examples in this repo are written the way that turns it off:

```rust
container()
    .height(300.0)
    .scrollable(ScrollAxis::Vertical)
    .child(
        container()                              // ← one child; the window
            .layout(Flex::column().spacing(10.0)) //   has nothing to narrow
            .children(keyed(..)),                 //   and the 10 000 rows are
    )                                             //   culled one at a time
```

## Observed

`src/widgets/container/mod.rs:1583-1624`. The window is gated on
`is_scrollable`, and its axis comes from `self.scroll_axis`:

```rust
let visible_children: &[WidgetId] = if is_scrollable {
    match self.scroll_axis {
        ScrollAxis::Vertical => { /* two partition_points over all_children */ }
```

A wrapper is one child, so the slice is the whole of it, and the 10 000 rows
below it reach `paint_children` unnarrowed.

### Measured

`examples/perf_stress_test.rs`, 10 000 rows, release with `render-stats`, the
same scroll gesture both ways:

| per second | wrapped | rows as direct children |
|---|---|---|
| children examined | **960 384** | **1 275** |
| paint phase | 105 µs | **13 µs** |
| paint cache hit rate | 0.1 % | **49.5 %** |
| scroller window | 1 offered, 1 iterated | 10 000 offered, **471 iterated** |

753× fewer children examined and an 8× faster paint, from moving `.layout()`
onto the container that already had `.scrollable()`.

### One correction to the original report, and it makes this worse

"the 10 000 rows are culled one at a time" is true only of *clean* rows. The
per-child cull in `src/widgets/paint_children.rs:71-73` is guarded by
`!tree.needs_paint(child_id)`:

```rust
if let Some(ref cull) = opts.cull_rect
    && !tree.needs_paint(child_id)
```

So on a first paint — and after anything that dirties the subtree — **nothing is
culled at all** and every row is painted in full. The binary-search window is
the only mechanism that bounds work unconditionally; the cull is a
second-frame-onwards optimisation that rides on the paint cache. Wrapping does
not degrade virtualization to a linear scan, it removes the only unconditional
bound there is.

This is already committed in the harness. `tests/snapshots/scrolling.snap` is
the first scenario of `tests/render_snapshots.rs:284`, written in the wrapped
form, and it lists all twenty rows of a 200 px-tall vertical scroller —
including ones at `y=338`, `y=371`, `y=404`, far below the viewport:

```
node 0,0 200x200 at 8,8 clip(0,0 200x200 r=8)
  draw rect 0,0 200x200 fill=#262633ff radius=8/8/8/8
  node 0,0 136x648
    node 0,0 120x24 at 8,8
    ...
    node 0,0 120x24 at 8,338      ← 138 px below the viewport, painted
```

The snapshot has been recording the defect since it was blessed. Nothing asked
the question it answers, because **no test anywhere covers the window**:
`partition_point` and `record_scroll_paint_range` appear only in
`src/widgets/container/mod.rs` and `src/render_stats.rs`, and in no test file.

### The precondition nobody wrote down

The binary search needs the children sorted and non-overlapping along the axis.
Nothing checks that; it is inferred from `scroll_axis`. A container declared
`.scrollable(ScrollAxis::Vertical).layout(Flex::row())` binary-searches bounds
that are sorted along `x`, and paints the wrong window. That is latent today and
becomes load-bearing under any fix that widens where the window applies.

## Expected

A scrollable list is virtualized whether its rows are its direct children or sit
under a wrapper — or, failing that, the case where it is not says so instead of
reporting `1/1` and looking healthy.

## The decision

**C: the container observes the result of its own layout.** Agreed with the
maintainer before implementation; A and B were weighed and rejected, recorded
here so the same ground is not covered twice.

`Container::layout` already walks its children. After `self.layout.layout(..)`
returns, it records in one pass whether they came out ascending and
non-overlapping along an axis, and caches that `Option<Axis>` on the container.
`paint` windows whenever it has a cull rect and a cached axis — scrollable or
not, at any depth, with or without siblings.

```rust
// Container::layout, after the layout impl has run
self.sorted_axis = sorted_axis_of(tree, children);

// Container::paint — the window stops being about scrolling
let visible_children = match (effective_cull_rect, self.sorted_axis) {
    (Some(cull), Some(axis)) => window(all_children, tree, cull, axis),
    _ => all_children,
};
```

### Why C

Coverage is the dominant term, not the per-frame constant. At 60 fps the
measured numbers above are ~16 000 children examined per frame wrapped against
~21 direct.

| | per-frame paint cost | shapes fixed | cost elsewhere |
|---|---|---|---|
| **C** observe | a plain `Option<Axis>` field read — none added | any depth, any siblings | one O(n) bounds scan per *layout*, early-exit on the first unsorted pair |
| **A** trait method | virtual call + `Signal<Axis>::get()` per candidate container per frame | any depth, any siblings | none |
| **B** descend one level | slice match + one tree lookup | only a lone wrapper at depth 1 | none |

- **B loses on coverage, not on a constant.** A wrapper inside a wrapper, a
  wrapper with a sibling, a component that returns a column — all stay at
  ~16 000 per frame. The 753× is still there for anything that is not the exact
  shape of the three examples.
- **A and C cover the same shapes**, so it comes down to where the cost lands.
  C's is confined to layout, which does not run during a scroll gesture:
  scrolling writes `scroll_state.offset_y` and only paint reads it, so the hot
  path pays nothing. A's runs every frame for every container holding a cull
  rect, just to ask whether it can window.
- **A has a correctness wrinkle before it has a cost one.**
  `FlexLayout::direction` is a `Signal<Axis>` (`src/layout/flex_layout.rs:43`)
  read inside `with_signal_tracking(id, JobType::Layout, ..)`. Calling
  `stacking_axis()` from paint reads it outside that scope.
- **C's scan is cheaper than it sounds.** It runs immediately after the layout
  walked the same children, so the bounds are warm, and it bails at the first
  out-of-order pair — a `ZStack`, or a row inside a vertical scroller, exits at
  child two. The full pass is only paid where it earns the 750×.
- **C is the only one that closes the unwritten precondition.** Sortedness
  becomes measured rather than assumed, so
  `.scrollable(ScrollAxis::Vertical).layout(Flex::row())` stops binary-searching
  x-sorted bounds and paints the right window.

Against C, and accepted: it infers where every comparable toolkit declares (see
**Prior art**). Guido already infers, for direct children, and has since the
window was written — so this finishes an inference rather than starting one, and
the inference is checkable rather than a promise, because the window only ever
narrows to what the cull rect already says is visible.

### The counter is in scope

`record_scroll_paint_range` (`src/render_stats.rs:328`) holds both numbers
already and today reports `1/1` for a container that gave up — indistinguishable
from one whose window covered everything. A container that could not window
becomes distinguishable from one that did, and every windowing container
records, not only scrollable ones.

## Acceptance

Two tests, one new and one existing.

**New — `tests/scroll_virtualization.rs`.** A 200 px-tall
`ScrollAxis::Vertical` container over 200 rows of 24 px with 8 px spacing, laid
out and painted through the same harness `tests/render_snapshots.rs:47`
(`render`) uses — `Tree::new`, `register_children`, `layout`, `paint`, and count
the nodes in the resulting `RenderNode`. Written twice: rows as direct children,
and rows wrapped in `container().layout(Flex::column().spacing(8.0))`. Assert
that both forms paint the same, bounded number of rows — roughly viewport ÷ row
pitch, plus the one-either-side the existing `saturating_sub(1)` / `+ 1`
already adds.

Today the direct form paints ~10 and the wrapped form paints 200, so the
assertion fails on the wrapped form, and it fails for the stated reason: the
window saw one child.

**Existing — `tests/snapshots/scrolling.snap` moves, and that is the point.**
Its first scenario is the wrapped form. After the change it stops naming the
rows below `y=200`, and the second scenario (`column(2)`, everything visible) and
the horizontal one are unaffected except by their own windows. This is an
intended reference change, so it is `REBLESS_SNAPSHOTS=1` and the pull request
carries `golden-update`; the diff — rows outside a 200 px viewport disappearing
from a paint — *is* the review. It is not evidence of a regression and should
not be read as one.

**Third — the counter.** Under `--features render-stats`,
`scroll_children_total` / `scroll_children_iterated` for the wrapped form
reports the row count and the window rather than `1/1`, and a container that
declined to window is distinguishable from one whose window covered everything.
This is what puts a number on the 753× in the pull request.

## Non-goals

- **A lazy list widget.** No builder delegate, no recycling, no
  `LazyColumn`. Every row still exists as a widget and is still laid out; only
  what reaches paint is narrowed. That is a larger question and this change
  should not start it.
- **`ScrollAxis::Both`.** Still falls back to full iteration, as today.
- **The per-child cull in `paint_children`.** It is correct and complementary;
  this changes what reaches it, not what it does.
- **Rewriting the examples to the direct form.** That is the report's third
  option and this change is meant to make it unnecessary. Worth confirming
  afterwards that the examples got faster without being touched.
- **Horizontal-in-vertical nesting and other mixed-axis trees.** Whatever the
  chosen mechanism answers for them, it answers by declining to window, not by
  guessing.

## Where it lives

- `src/widgets/container/mod.rs:1561-1571` — `effective_cull_rect`, the only
  place the scrollable and non-scrollable paths already agree
- `src/widgets/container/mod.rs:1583-1624` — the window
- `src/widgets/container/mod.rs:1277-1360` — `Container::layout`, and the
  early-out that a cached axis would have to survive
- `src/widgets/paint_children.rs:55-115` — the per-child cull and how the cull
  rect reaches a child's coordinate space
- `src/layout/mod.rs:276` — `Layout`, public and one method
- `src/layout/flex_layout.rs:43,409`, `src/layout/zstack.rs:73`
- `src/render_stats.rs:328` — `record_scroll_paint_range`
- `tests/render_snapshots.rs:284`, `tests/snapshots/scrolling.snap`
- `examples/bench_list.rs:35`, `examples/scroll_example.rs:27`,
  `examples/perf_stress_test.rs:76` — all three wrapped

## Prior art

The question guido has not settled is whether laziness is **declared** or
**inferred**. Every comparable toolkit declares it, and none of them virtualizes
a plain box nested inside a scroll area:

- **Flutter** — [`ListView`](https://api.flutter.dev/flutter/widgets/ListView-class.html)
  with a [`SliverChildBuilderDelegate`](https://api.flutter.dev/flutter/widgets/SliverChildBuilderDelegate-class.html)
  builds children on demand; `RenderSliverList` asks the delegate for indices as
  the viewport moves. A `Column` inside a `ListView` is one sliver child and is
  built whole — the same defect, and Flutter's answer is that the lazy thing is
  a different widget, not an optimisation applied to an ordinary one.
- **SwiftUI** — [`LazyVStack`](https://developer.apple.com/documentation/swiftui/lazyvstack)
  against `VStack`. Same shape: laziness is a property of the stack type the
  author chose.
- **GTK 4** — [`GtkListView`](https://docs.gtk.org/gtk4/class.ListView.html)
  over a `GListModel` with a factory that recycles rows. A `GtkBox` inside a
  `GtkScrolledWindow` does not.
- **Qt** — `QListView`'s model/view against `QScrollArea` plus a layout. Again
  structural.
- **CSS / Chromium** — [`content-visibility: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility)
  with `contain-intrinsic-size` is the nearest thing to "make it not matter at
  any depth": the engine skips layout and paint of a subtree that is offscreen.
  It is still **declared per element**, on the element that should be skipped,
  and it needs an intrinsic size because skipping layout loses the geometry.
  Notably a plain `overflow: scroll` box virtualizes nothing.

What that says about the decision taken: nobody infers, so **C** is the one with
no precedent. It is taken anyway, because guido already infers — for direct
children, and has since the window was written. The choice was never between
inferring and declaring; it was between making the inference guido already makes
work properly and at depth, or leaving it partial and documenting the shape.
That last is the report's third option, which every toolkit above effectively
took, and which the report itself argues does not survive someone not reading
the docs.

CSS's `contain-intrinsic-size` is also a warning about how far this can go: the
moment a mechanism skips *layout* rather than paint, the sizes have to come from
somewhere. This change only narrows paint, which is why it needs no such
declaration — worth saying out loud so a later change does not walk into it.

MalpenZibo:
## What changed

The binary search that narrows a container's children to the ones the viewport
can reach used to run only when the container was scrollable, and took its axis
from `scroll_axis`. Wrapping the rows in a layout container — which is how all
three scrollable examples in this repo are written — left the scroller with one
child to narrow, and turned virtualization off completely with no warning and a
counter reading `1/1`.

The window was never about scrolling. It needs a rect to narrow to and children
ordered along an axis to search; a scroller supplies the first from its own
viewport and everything below it inherits one. So it moves into
`paint_children`, which already owns narrowing children to a cull rect and
belongs to no widget in particular, and it applies wherever both halves are
present — at any depth, with or without siblings.

The axis is measured, not declared. `Container::layout` records, in the pass
that just laid the children out, whether they came out non-overlapping along an
axis — which is exactly what `partition_point` needs of the slice and what
nothing had ever checked.

Closes #204.

## What proves it

**`tests/scroll_virtualization.rs`, three tests.** Confirmed red on
`origin/main` and green here — the reviewer re-checked this independently by
copying the file onto 5f2bc37.

- `a_wrapped_list_paints_the_rows_an_unwrapped_one_paints` — 200 rows in a
  200 px scroller, written both ways. **200 painted wrapped against 8 direct**
  before; equal after.
- `a_scroller_whose_children_are_ordered_along_neither_axis_paints_all_of_them`
  — this one is a **correctness** fix, not a performance one. `partition_point`
  over an unpartitioned slice returns whichever index it probed. Eight rows of
  which two are in the viewport: the old search painted the first and dropped
  the last. **2 of 8** before, 8 of 8 after.
- `the_counter_separates_a_narrowed_window_from_one_that_could_not_narrow` —
  under `--features render-stats`, over a real paint.

**Two snapshots moved, and the PR carries `golden-update`.**
`tests/snapshots/scrolling.snap`: vertical scroller 20 → 7 rows, horizontal
10 → 4, and the middle scroller whose content fits is unchanged and *not*
marked partial. `tests/snapshots/bar_like_composition.snap`: 6 → 4 swatches.
Every dropped node is outside its viewport; the three new `partial` markers are
the narrowed paints. The reviewer recomputed both by hand against the window
arithmetic and they match. The re-bless was the maintainer's decision, taken
with this diff in front of them.

No golden moved, and none should: no golden scenario holds a scroller, and the
table in `AGENTS.md` puts "what gets drawn and where" on the render-tree
snapshots.

## The measurement

`examples/perf_stress_test.rs` — 10 000 rows, already written in the wrapped
form — release with `render-stats`, the same scroll gesture both ways.
Normalised per *painted* frame, since the two runs saw different amounts of
scrolling:

| per painted frame | before | after |
| --- | --- | --- |
| children examined | **10 004** | **12** |
| paint phase, avg | **174–187 µs** | **12–16 µs** |
| paint cache hit rate | **0.1 %** | **46–52 %** |
| the window itself | `scroll: total_children=1 iterated=1` | `window: offered=10 002 narrowed_to=7.6 \| declined=0` |

Raw, one representative block each:

```
before   paint: children=530212 cached=308 painted=106 culled=529798 cache_rate=0.1%
         timing paint=174/92/505 us
         scroll: total_children=53 iterated=53          (53 painted frames)

after    paint: children=179 cached=86 painted=50 culled=43 cache_rate=48.0%
         timing paint=12/8/26 us
         window: offered=150030 narrowed_to=114 | declined=0 in 0 containers   (15 painted frames)
```

**The predicted cost did not appear.** The review flagged that
`visible_window` calls `ctx.mark_partial()` where the old scroll window never
did, so a windowing container and its ancestors become uncacheable — a cost
pointing the other way. Measured, the paint cache went **up**, 0.1 % → 48 %:
the wrapper was already partial every frame from the per-child cull, and what
the window buys the ten thousand rows underneath it dwarfs one uncacheable
node.

## What the review found

The `reviewer` subagent read the commit before this existed. **Zero blocking.**
Three worth answering, all acted on:

1. **No after-measurement.** Fixed — the table above. This is the finding that
   mattered; the numbers in the issue were taken on the old code comparing
   wrapped against a hand-restructured tree, and nothing measured this branch.
2. **Nothing exercised the counters through the paint path.** Added
   `the_counter_separates_a_narrowed_window_from_one_that_could_not_narrow`.
   Writing it found a real bug: every childless leaf under a scroller was being
   counted as a window that could not narrow, which would have buried the one
   container that matters. The nothing-to-narrow cases are now excluded before
   the counting rather than after.
3. **`ScrollAxis::Both` was a stated non-goal and no longer holds.** True, and
   flagged here rather than left for a reader to discover. The `_ => all_children`
   arm is gone, so a `Both` scroller whose children come out ordered now
   windows. It is correct — non-overlap along the searched axis is all the
   search needs, whatever the other axis does — and free.

Notes acted on: `docs/ARCHITECTURE.md` said `partial` meant "children were
culled", the twin of the `docs/RENDERER.md` line already widened; `sorted_axis`
answered `Vertical` for an empty child list, a claim about nothing; the comment
on the one-child-either-side slack implied it bounds a transform, which it does
not.

One note not acted on: the reviewer would have accepted a split into two
commits and did not object to one. I kept one — a first commit measuring an
axis nothing reads is not obviously better, and the counter rename really is
part of "the window is not about scrolling".

## What was checked by hand

`examples/perf_stress_test` on niri, before and after, driven by the
maintainer — the measurement above. The list scrolls, the rows are the right
rows, and nothing flickers or comes back at the top, which is the failure mode
`partial` exists to prevent.

## Left undone

**#316** — a cull rect is translated into a child's coordinate space by the
child's laid-out origin and never composed with the child's own transform, so a
transformed child is culled against where it *would* have been. Older than this
change and unchanged by it in steady state; what this spends is one frame, the
first, where the per-child cull had not started yet and `main` was accidentally
right. Found by the review, filed rather than fixed: it is not this change's
defect and it needs its own tests.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -554,7 +554,7 @@ The paint system tracks which widgets need repainting:
   subsequent frames, a clean child whose position didn't change is reused via `Rc::clone`
   (zero copies); if it moved, only the node header is cloned (children and commands stay
   shared) with the position recomposed from the decomposed parent/user transforms.
-- **Partial propagation**: a node whose children were culled (`partial`) poisons its
+- **Partial propagation**: a node that did not paint all of itself (`partial`) poisons its
   ancestors in the cache walk — incomplete paints are never cached, so reuse can never
   resurrect a subtree with missing children. Refusing to cache is only half of it: a
   partial paint also *drops* the entry the last complete one left, because the widget
--- a/docs/RENDERER.md
+++ b/docs/RENDERER.md
@@ -31,7 +31,7 @@ pub struct RenderNode {
     pub clip: Option<ClipRegion>,            // Clips this node and children
     pub overlay_clip: Option<ClipRegion>,    // Clips only overlay commands
     pub repainted: Cell<bool>,               // true = freshly painted this frame
-    pub partial: bool,                       // Some children were culled (do not cache)
+    pub partial: bool,                       // Some children were not painted (do not cache)
     pub cached_flatten: RefCell<Option<Rc<CachedFlatten>>>, // Cached flatten output
 }
 ```
@@ -49,7 +49,8 @@ to the same node that sits in the frame's render tree:
   cached nodes are shared: `repainted` is cleared once the node's output has
   been cached, and flatten writes `cached_flatten` through a `&RenderNode`.
 - `partial` propagates to ancestors during the cache walk: a subtree that
-  embeds a partially-painted node (culled children) is never cached, so a
+  embeds a partially-painted node — children narrowed away by the visible
+  window, or culled one at a time beneath it — is never cached, so a
   later cache reuse cannot resurrect an incomplete paint. It also invalidates:
   `Tree::clear_cached_paint` drops whatever the last complete paint left, which
   is a picture of the widget as it no longer is — a list that scrolls culls
--- a/src/render_stats.rs
+++ b/src/render_stats.rs
@@ -62,9 +62,12 @@ pub struct StatsSnapshot {
     pub flatten_timing: PhaseTiming,
     pub gpu_render_timing: PhaseTiming,
     pub cache_paint_timing: PhaseTiming,
-    // Scroll
-    pub scroll_children_total: u64,
-    pub scroll_children_iterated: u64,
+    // The paint window: children offered to it, children it let through, and
+    // the ones it could not narrow because nothing ordered them.
+    pub window_children_total: u64,
+    pub window_children_iterated: u64,
+    pub window_declined_children: u64,
+    pub window_declined_containers: u64,
 }
 
 /// Zero-cost timing macro. Wraps a block with `Instant::now()` / `.elapsed()`
@@ -166,9 +169,11 @@ mod inner {
         flatten_phase: PhaseAccum,
         gpu_render_phase: PhaseAccum,
         cache_paint_phase: PhaseAccum,
-        // Scroll
-        scroll_children_total: u64,
-        scroll_children_iterated: u64,
+        // Paint window
+        window_children_total: u64,
+        window_children_iterated: u64,
+        window_declined_children: u64,
+        window_declined_containers: u64,
         // Report timing
         last_print: Instant,
     }
@@ -195,8 +200,10 @@ mod inner {
                 flatten_phase: PhaseAccum::new(),
                 gpu_render_phase: PhaseAccum::new(),
                 cache_paint_phase: PhaseAccum::new(),
-                scroll_children_total: 0,
-                scroll_children_iterated: 0,
+                window_children_total: 0,
+                window_children_iterated: 0,
+                window_declined_children: 0,
+                window_declined_containers: 0,
                 last_print: Instant::now(),
             }
         }
@@ -221,8 +228,10 @@ mod inner {
             self.flatten_phase.reset();
             self.gpu_render_phase.reset();
             self.cache_paint_phase.reset();
-            self.scroll_children_total = 0;
-            self.scroll_children_iterated = 0;
+            self.window_children_total = 0;
+            self.window_children_iterated = 0;
+            self.window_declined_children = 0;
+            self.window_declined_containers = 0;
             self.last_print = Instant::now();
         }
     }
@@ -323,13 +332,28 @@ mod inner {
         });
     }
 
-    /// Record scroll paint iteration stats.
+    /// Record a container that narrowed its children to the visible rect.
     #[inline]
-    pub fn record_scroll_paint_range(total_children: u64, iterated: u64) {
+    pub fn record_paint_window(total_children: u64, iterated: u64) {
         STATS.with(|s| {
             let mut stats = s.borrow_mut();
-            stats.scroll_children_total += total_children;
-            stats.scroll_children_iterated += iterated;
+            stats.window_children_total += total_children;
+            stats.window_children_iterated += iterated;
+        });
+    }
+
+    /// Record a container that had a rect to narrow to and could not: its
+    /// children are ordered along no axis, so every one of them is examined.
+    ///
+    /// This is the number worth watching. A container in this state looks
+    /// healthy from the outside — it draws the right thing — and pays for the
+    /// whole of its list on every frame that repaints it.
+    #[inline]
+    pub fn record_paint_window_declined(total_children: u64) {
+        STATS.with(|s| {
+            let mut stats = s.borrow_mut();
+            stats.window_declined_children += total_children;
+            stats.window_declined_containers += 1;
         });
     }
 
@@ -357,8 +381,10 @@ mod inner {
                 flatten_timing: stats.flatten_phase.to_timing(),
                 gpu_render_timing: stats.gpu_render_phase.to_timing(),
                 cache_paint_timing: stats.cache_paint_phase.to_timing(),
-                scroll_children_total: stats.scroll_children_total,
-                scroll_children_iterated: stats.scroll_children_iterated,
+                window_children_total: stats.window_children_total,
+                window_children_iterated: stats.window_children_iterated,
+                window_declined_children: stats.window_declined_children,
+                window_declined_containers: stats.window_declined_containers,
             }
         })
     }
@@ -458,11 +484,14 @@ mod inner {
                     ct.avg_us, ct.min_us, ct.max_us,
                 );
 
-                // Scroll stats (only if scroll activity occurred)
-                if stats.scroll_children_total > 0 {
+                // The paint window, and what it could not narrow.
+                if stats.window_children_total > 0 || stats.window_declined_children > 0 {
                     eprintln!(
-                        "  scroll: total_children={} iterated={}",
-                        stats.scroll_children_total, stats.scroll_children_iterated
+                        "  window: offered={} narrowed_to={} | declined={} children in {} containers",
+                        stats.window_children_total,
+                        stats.window_children_iterated,
+                        stats.window_declined_children,
+                        stats.window_declined_containers
                     );
                 }
 
@@ -529,7 +558,11 @@ pub fn record_phase_duration(_phase: Phase, _duration: std::time::Duration) {}
 
 #[cfg(not(feature = "render-stats"))]
 #[inline(always)]
-pub fn record_scroll_paint_range(_total_children: u64, _iterated: u64) {}
+pub fn record_paint_window(_total_children: u64, _iterated: u64) {}
+
+#[cfg(not(feature = "render-stats"))]
+#[inline(always)]
+pub fn record_paint_window_declined(_total_children: u64) {}
 
 #[cfg(not(feature = "render-stats"))]
 #[inline(always)]
@@ -738,12 +771,25 @@ mod tests {
     }
 
     #[test]
-    fn test_scroll_paint_range() {
+    fn test_paint_window() {
         setup();
-        record_scroll_paint_range(10000, 52);
-        record_scroll_paint_range(10000, 48);
+        record_paint_window(10000, 52);
+        record_paint_window(10000, 48);
         let s = get_stats();
-        assert_eq!(s.scroll_children_total, 20000);
-        assert_eq!(s.scroll_children_iterated, 100);
+        assert_eq!(s.window_children_total, 20000);
+        assert_eq!(s.window_children_iterated, 100);
+    }
+
+    /// A container that could not narrow is counted apart from one whose
+    /// window happened to cover everything. Both used to read as a window.
+    #[test]
+    fn test_paint_window_declined() {
+        setup();
+        record_paint_window(10000, 10000);
+        record_paint_window_declined(10000);
+        let s = get_stats();
+        assert_eq!(s.window_children_iterated, 10000);
+        assert_eq!(s.window_declined_children, 10000);
+        assert_eq!(s.window_declined_containers, 1);
     }
 }
--- a/src/widgets/container/mod.rs
+++ b/src/widgets/container/mod.rs
@@ -26,7 +26,7 @@ use crate::advance_anim;
 use crate::animation::{Animatable, IntoAnimated, Motion};
 use crate::backdrop::BackdropBlur;
 use crate::jobs::{JobRequest, JobType, RequiredJob, request_job};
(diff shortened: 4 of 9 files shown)


MalpenZibo:
## What the mutation job found

Red, and read rather than merged past. The first run reported **23 missed, 36 caught**; after the second commit it is **22 missed, 37 caught**, and the one that changed is the one that mattered.

**One real gap, now closed.** `cargo mutants` turned `horizontal &= left >= right` into `|=` in `sorted_axis` and nothing objected. The vertical half was watched and the horizontal half was not, because the scattered rows in the test all sat at one `x` — so a wrong horizontal answer moved no row, and the function could claim the children were ordered along an axis they were not. The rows now scatter on both axes in the same shape. Confirmed by applying both mutants by hand: each fails two of the three tests, where the horizontal one used to fail none. It is absent from the second run's survivors.

**Twenty are unreachable rather than unwatched.** All in `render_stats.rs`, whose `inner` module is entirely behind `#[cfg(feature = "render-stats")]`. The job runs `-C --lib -C --tests` with no `--all-features`, so that module is never compiled and no test can kill a mutant inside it — including `the_counter_separates_a_narrowed_window_from_one_that_could_not_narrow`, which is itself `#[cfg(feature = "render-stats")]` and does kill them under `cargo test --all-features`. This is the same shape as the doc tests the job's own comment records losing, and it is not specific to this change: any diff touching feature-gated code will report it as unwatched. Worth an issue against the workflow if anyone wants that number to mean what it looks like — say so and I will open one.

**Two are the `children.len() < 2` boundary** at `paint_children.rs:91`. `== 2` is equivalent in a build without the feature: for one child the window returns the whole slice and marks nothing partial, and for none there is nothing to return. `<= 2` differs only for a two-child list scrolled entirely past the viewport, where the one-either-side slack brings the answer back to both children anyway. Left, with the reasoning here rather than a test contorted around them.

The job reports and does not block, per `AGENTS.md`. Every blocking check is green.

Issue #319, opened 2026-09-02
A widget culled before it ever painted cannot bring itself back into view


MalpenZibo:
## Observed

A widget that has never painted holds no paint subscriptions, so nothing can
wake it. Once culled, it stays culled however loudly its own signals change.

The reactive system registers a subscriber when a widget *reads* a signal, and a
widget reads its paint properties inside its own `paint`. A widget culled before
it ever painted has read nothing, so a write that would bring it on screen
notifies no one:

```
20-row scroller, row 15 never painted (outside the window from the first frame)
lift.set(-400.0)          // would carry it into the viewport
jobs::queued_job_types(row)  => []
jobs::queued_job_types(root) => []
paint_overflow(row)          => 0
```

The row is not painted, and nothing has been asked to reconsider.

Verified against `main` as well: it culls the same row for the same reason, so
this is not something #316 introduced. What #316 does is make the rest of that
sentence true — a widget that *has* painted now refreshes its reach from the
Paint job its own write queues — which is what makes the exception worth naming.

## Why it is not obviously wrong

A widget outside the window has nothing on screen to update, and the cheapest
correct answer for "how far can this move" is the one it gave when it last
painted. The alternative — subscribing to a widget the frame never touches —
is what culling exists to avoid.

But it means the guarantee is narrower than the reach machinery's doc implies:
`Container::max_transform_reach` says the value is kept current by the schedule,
and that holds only for a widget that has painted at least once.

## Expected

Either a widget culled from the first frame can still bring itself back into
view with a transform, or the docs say plainly that it cannot and why.

## Acceptance

The cheap half first, because it can be done today and the other half is a
design question: a line in `Container::max_transform_reach`'s doc naming the
exception, and — if the answer is that this is intended — a test in
`src/widgets/container/characterization.rs` pinning it, so the next reader finds
a decision rather than a gap.

If the answer is that it should wake: a test in the shape of
`tests/cull_follows_the_transform.rs` where a row that has never been painted is
given a transform bringing it into the viewport, and is painted on the next
frame. It fails today, on this branch and on `main`.

## Non-goals

- **Culling itself.** Not painting what cannot be seen is the point; this is
  about what happens when that judgement stops being true.
- **The window's arithmetic**, which #316 settles.

## Where it lives

- `src/widgets/container/style.rs` — `max_transform_reach`, and the paragraph
  about what keeps its answer current
- `src/widgets/container/mod.rs` — `refresh_paint_bounds`, which runs from a
[...]

Issue #318, opened 2026-09-02
Damage covers where a widget's paint went, never where it came back from


MalpenZibo:
## Observed

Damage covers where a widget's paint *went*, never where it came *back from*.

`Tree::surface_relative_bounds_and_root` grows a widget's damage rect by its
`paint_overflow` — how far it paints outside its own bounds — and that value is
refreshed before the rect is computed. So the rect always describes the widget
as it is *now*, and the pixels it occupied a moment ago are never named.

Measured on a 50x50 child inside a 200x200 root:

| the child | damage reported |
| --- | --- |
| moves from rest to `translate(0, +100)` | `-90,-90 250x250` — covers where it went |
| moves back to rest | `10,10 50x50` — the pixels at y=110..160 are never damaged |

The buffer is redrawn either way; `wl_surface.damage_buffer` is what decides
what the compositor re-composites, so what is left is a fringe of the widget at
its old position.

## It is older than the transform work

Found by the review of #316, which is what made half of it visible. On `main`
this shape damages `10,10 50x50` in **both** directions, because a transform
contributed nothing to `paint_overflow` at all — the outbound move was as
uncovered as the return. #316 closes the outbound half and leaves this one.

The general form is wider than transforms: **nothing damages the rect a widget
vacates when any paint-only property shrinks its reach.** A shadow that animates
from elevation 8 to 0 is the same shape, and predates all of this.

`Tree::set_origin` and `Tree::cache_layout` already do the right thing for the
geometry they own — they damage the old rect before the new one takes effect.
The paint-property path has no equivalent.

## Expected

A widget whose reach shrinks damages what it is vacating as well as what it now
covers, so a compositor re-composites both.

## Acceptance

Nothing tests damage against a transform at all today, so the scenario is the
work. In `src/tree.rs`'s own tests or beside `tests/paint_cache_across_frames.rs`,
which is where cross-frame behaviour lives: a child translated away from rest
and then back, asserting the damage rect on the returning frame contains the
rect it occupied on the frame before. It fails today with the numbers above.

Worth writing the elevation twin beside it, so the fix is not made transform-
shaped when the defect is not.

## Non-goals

- **The one-scalar reach.** `paint_overflow` is symmetric, so a widget that
  moves down widens its damage upward too. Deliberate, and a separate question
  from remembering where it was.
- **Damage for a subtree below a moved widget**, which is
  `surface_relative_bounds_and_root` summing ancestor origins and ignoring
[...]

PR #320, opened 2026-09-02
A widget is culled where it draws, not where it was laid out

### LINKED ISSUE #316: A cull rect follows a child to where it was laid out, not to where its transform draws it

MalpenZibo:
## What changed

Two things narrow a container's children to a cull rect — the binary-search
window and the per-child cull beside it — and both asked about **laid-out**
bounds. A child's own `translate`, `rotate` or `scale` is applied inside its own
`paint`, after the parent has already decided, so a subtree lifted 300px was
still chosen by where it would have been. A scroller over a lifted column
painted eight rows and every one of them sat above the viewport.

Now the rect follows the transform as it descends, and a widget says how far its
paint can land outside its box so whoever is about to narrow it knows to keep
it. **A transform is still paint work**: nothing about this reflows anything.

Closes #316.

## What proves it

`tests/cull_follows_the_transform.rs` — five scenarios, all red before:

- a lifted column paints the rows the lift brings into view, not the rows that
  would have been there
- an *un*lifted column still narrows, so the fix cannot be "widen until nothing
  is culled"
- a row its own transform lifts into view survives **both** frames — the window
  and the per-child cull take turns, and only the second frame reaches the cull
- a row whose **child** is lifted is kept: the reach gathers upward, or the row
  is culled at its laid-out bounds and the visible box goes with it
- a subtree scaled to nothing paints nothing: `scale(0.0)` has no inverse, and
  the answer is the empty rect rather than *no* rect. **20 rows** are painted
  without that, measured.

`mod reach_tests` in `src/widgets/container/style.rs` pins the arithmetic, which
a review found ten pixels short by working an example by hand: the shadow and
the transform compose rather than alternate, and a rotation's outset is not
monotone in its angle, so it gets the worst angle outright — bounded by the
circle its furthest corner sweeps about the pivot, which a centre-shaped bound
under-reports by 50px for a corner pivot.

In `src/widgets/container/characterization.rs`, the invalidation contract:

| test | mechanism | with it removed |
| --- | --- | --- |
| `a_transform_invalidates_paint_only` | translate/rotate/scale queue `[Paint]`, not `[Layout]` | — |
| `a_transform_does_not_reflow_the_parent_either` | the reads are untracked | `got [Layout]` |
| `a_declared_transform_change_invalidates_the_reach_without_a_layout` | the `refresh_paint_bounds` hook | `left: 0.0, right: 30.0` |
| `a_reach_that_grows_under_a_relayout_boundary_still_reaches_the_parent` | `gather_reach_upward` | `left: 0.0, right: 70.0` |
| `a_reach_that_shrinks_lets_the_parent_narrow_again` | the shrink path re-measures | `left: 70.0, right: 0.0` |

Those right-hand numbers are runs, not claims: each mechanism was deleted in
turn and the test watched to fail. An earlier revision of this branch cited two
of these tests in its commit message when they had been destroyed by a stray
`git checkout` — the review caught it by mutation, which is why they are
tabulated here with what actually happens.

`tests/snapshots/scrolling.snap` gains one row, re-blessed with the maintainer's
authorisation — hence `golden-update`. Its rows carry a one-pixel translate
each, so the widest reach is 19px, the window widens by 19px on a 33px pitch,
and one more row falls inside. **One row rather than the thirteen** an
all-or-nothing answer costs — which is what an earlier attempt did, and why the
reach is a magnitude rather than a boolean.

No golden moved, and none should: no golden scenario holds a scroller.

## Prior art

The mechanism was agreed with the maintainer before it was written, after
reading up how others cull a transformed child.

- **Blink** runs a [PrePaint tree walk](https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/core/paint/README.md)
  between layout and paint that builds the transform property trees and computes
  cull rects, driven by `NeedsPaintPropertyUpdate` — *paint*-level dirty bits,
  never layout ones. It culls against the **visual rect**, "the bounding box of
  all pixels that will be painted", not the layout box.
- **Flutter** gives every `RenderObject` an
  [`applyPaintTransform`](https://api.flutter.dev/flutter/rendering/RenderObject/applyPaintTransform.html)
  and a [`paintsChild`](https://api.flutter.dev/flutter/rendering/RenderObject/paintsChild.html),
  so a parent can ask a child where it will paint *before* painting it. Both
  Flutter and Blink union descendant paint bounds up the tree, which is what
  `gather_reach_upward` does.
- **Qt** indexes `QGraphicsScene` on `boundingRect` and culls against it, with
  the item's transform already applied.
- **GTK4** puts the transform in the render node tree, so a `GskTransformNode`'s
  bounds are its child's run through `gsk_transform_transform_bounds`.

The common answer: **cull against geometry that already accounts for the node's
own transform, computed in a pass after layout and before paint, invalidated by
paint-level dirty bits.** Guido already had that pass — `process_jobs` — so this
adds `Widget::refresh_paint_bounds` (defaulted, so no existing implementor
breaks) and calls it from the Paint loop. Layout publishes the value too,
because the reach depends on bounds and only layout knows those, but nothing
subscribes it.

Rejected on the way, recorded so the ground is not covered twice: a structural
"does this child declare a transform" flag, which cost a 20-row list its entire
window for a decorative 1px translate; and reading the reach under
`JobType::Layout`, which made a transform a layout property — the one thing it
must not be.

## What the review found

Three passes. The first two returned **three blocking findings each round**, all
acted on; the third returned **one**, also acted on. The one that mattered most:
`snapshot_zone` does *not* suspend tracking — it silences the non-reactive-read
diagnostic and leaves the read registered — so an earlier revision claimed
transforms were paint-only while they were reflowing the *parent*. The test that
was supposed to prove it could not, because its container was the root, where
there is no enclosing scope to catch the read.

Findings answered rather than fixed:

- **The reach is one scalar for four sides**, so a widget that moves down widens
  the search and the damage rect in every direction. Deliberate; a directional
  reach would change `Tree::set_paint_overflow`'s signature and wants its own
  change.
- **`children_reach` is a shared max across siblings**, so one row lifted 400px
  widens the search for the whole list. Correct, and it costs virtualization in
  that case. The per-child cull already uses each child's own reach; only the
  binary search needs a single number, because a per-child reach makes its
  predicate non-monotone and there is nothing left to bisect.
- **A `.timeline(..)` is not bounded** by `max_transform_reach` — its keyframes
  are the animation's, not a declaration. Safe because a playing animation
  queues a Paint job per changed frame and the reach is recomputed from it; the
  doc says that now instead of claiming a bound it does not have.

## What was checked by hand

Nothing. Every claim above is a test in this branch, and the mechanisms are
watched by deleting them and running the suite. The full harness, `cargo clippy
--all-targets --all-features -- -D warnings`, `cargo test --test golden_images`
on lavapipe with `GUIDO_GPU_REQUIRED=1`, `documentation_references` and `mdbook
test` are all green.

## Left undone

- **#318** — damage covers where a transform went, never where it came back
  from. Measured: moving a child +100 damages `-90,-90 250x250`; moving it back
  damages `10,10 50x50` and the vacated pixels are never named. `main` gets
  *both* directions wrong, so this branch fixes half of it. The general form is
  wider than transforms — nothing damages what a widget vacates when any
  paint-only property shrinks its reach.
- **#319** — a widget culled before it ever painted holds no paint subscription,
  so a transform that would bring it back on screen notifies nobody. Pre-existing
  and unchanged here; what this branch adds is that a widget which *has* painted
  now refreshes correctly, which is what makes the exception worth naming.

### DIFF
--- a/.claude/skills/widgets/SKILL.md
+++ b/.claude/skills/widgets/SKILL.md
@@ -70,6 +70,14 @@ Two methods are required, the rest default:
 - `event(&mut self, tree, id, event) -> EventResponse` — defaults to `Ignored`
 - `advance_animations`, `reconcile_children`, `layout_hints`,
   `register_children` — defaults
+- `refresh_paint_bounds` — a default too, and the one a widget that transforms
+  itself has to override: a parent narrows its children to the visible rect
+  before painting them, by their laid-out bounds, so a widget that draws
+  somewhere else says how far with `Tree::set_own_paint_reach`. Called from the
+  Paint job, which is where a reach that follows a paint-only property belongs —
+  a transform must not reflow anything. A reach that follows a layout-tracked
+  signal is published from `layout` instead, as `Text` and `TextInput` do for
+  their stroke and shadow
 
 Position and bounds live in the `Tree`, never on the widget. Writing a widget
 from outside the crate needs `guido::widget_prelude::*` alongside the ordinary
--- a/book/src/advanced/custom-widgets.md
+++ b/book/src/advanced/custom-widgets.md
@@ -34,7 +34,20 @@ application never names `Transform`: it declares `translate`, `rotate` and
 
 `Widget` has two required methods. Everything else — `event`,
 `advance_animations`, `reconcile_children`, `layout_hints`,
-`register_children` — has a default.
+`register_children`, `refresh_paint_bounds` — has a default.
+
+One of those defaults is worth knowing about if your widget draws outside the
+box it was given. A parent narrows its children to the visible region before
+painting them, and it does that by their laid-out bounds — so a widget that
+paints elsewhere, because it transforms itself or casts something past its
+edges, has to say how far. That is `refresh_paint_bounds`, which reports it with
+`Tree::set_own_paint_reach`. It runs from the paint job rather than from layout,
+so saying it never costs a reflow.
+
+Which pass you publish from follows what your reach depends on. If it moves with
+something layout already tracks — a size, a font — publish it from `layout`,
+which is what the built-in text widgets do for their stroke and shadow. If it
+moves with a paint-only property such as a transform, publish it here.
 
 ```rust,ignore
 struct Bar {
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -414,6 +414,10 @@ pub trait Widget {
     /// Reconcile dynamic children. Returns true if children changed.
     fn reconcile_children(&mut self, tree: &mut Tree, id: WidgetId) -> bool { false }
 
+    /// Publish how far this widget's paint lands outside its bounds, before
+    /// anything decides whether to paint it. Called from the Paint job.
+    fn refresh_paint_bounds(&self, tree: &mut Tree, id: WidgetId) {}
+
     fn layout(&mut self, tree: &mut Tree, id: WidgetId, constraints: Constraints) -> Size;
     fn paint(&self, tree: &Tree, id: WidgetId, ctx: &mut PaintContext);
     fn event(&mut self, tree: &mut Tree, id: WidgetId, event: &Event) -> EventResponse;
--- a/src/jobs.rs
+++ b/src/jobs.rs
@@ -503,6 +503,12 @@ pub fn process_jobs(jobs: &[Job], tree: &mut Tree, layout_roots: &mut Vec<Widget
         }
     }
     for id in paint {
+        // Before the flag, and before this frame's paint: a parent decides
+        // which children to paint at the top of its own, so a widget that has
+        // moved has to have said so by now. See `Widget::refresh_paint_bounds`.
+        tree.with_widget_mut(id, |widget, wid, tree| {
+            widget.refresh_paint_bounds(tree, wid);
+        });
         tree.mark_needs_paint(id);
     }
     for id in layout {
--- a/src/renderer/flatten.rs
+++ b/src/renderer/flatten.rs
@@ -94,26 +94,7 @@ impl FlattenedCommand {
 
     /// The axis-aligned world box containing a rect of this command's own space.
     fn world_aabb(&self, rect: Rect) -> Rect {
-        let corners = [
-            self.world_transform.transform_point(rect.x, rect.y),
-            self.world_transform
-                .transform_point(rect.x + rect.width, rect.y),
-            self.world_transform
-                .transform_point(rect.x, rect.y + rect.height),
-            self.world_transform
-                .transform_point(rect.x + rect.width, rect.y + rect.height),
-        ];
-        let min_x = corners.iter().map(|c| c.0).fold(f32::INFINITY, f32::min);
-        let max_x = corners
-            .iter()
-            .map(|c| c.0)
-            .fold(f32::NEG_INFINITY, f32::max);
-        let min_y = corners.iter().map(|c| c.1).fold(f32::INFINITY, f32::min);
-        let max_y = corners
-            .iter()
-            .map(|c| c.1)
-            .fold(f32::NEG_INFINITY, f32::max);
-        Rect::new(min_x, min_y, max_x - min_x, max_y - min_y)
+        self.world_transform.map_rect(rect)
     }
 
     /// [`world_rounded_rect`](Self::world_rounded_rect), narrowed to what the
@@ -545,11 +526,11 @@ fn flatten_node(
     // Compute this node's world transform
     let (origin_x, origin_y) = node.pivot.resolve(node.bounds);
 
-    // Compose transforms: parent first, then local centered at origin
+    // Compose transforms: parent first, then local about its own pivot
     let local_centered = if node.local_transform.is_identity() {
         Transform::IDENTITY
     } else {
-        node.local_transform.center_at(origin_x, origin_y)
+        node.local_transform.about(node.pivot, node.bounds)
     };
     let world_transform = parent_world_transform.then(&local_centered);
 
@@ -790,45 +771,16 @@ fn world_bounds(cmd: &FlattenedCommand) -> Option<Rect> {
     ))
 }
 
-/// Compute axis-aligned bounding box from an array of points.
-fn aabb_from_points(points: &[(f32, f32)]) -> Rect {
-    let (min_x, max_x, min_y, max_y) = points.iter().fold(
-        (
-            f32::INFINITY,
-            f32::NEG_INFINITY,
-            f32::INFINITY,
-            f32::NEG_INFINITY,
-        ),
-        |(min_x, max_x, min_y, max_y), &(x, y)| {
-            (min_x.min(x), max_x.max(x), min_y.min(y), max_y.max(y))
-        },
-    );
-    Rect::new(min_x, min_y, max_x - min_x, max_y - min_y)
-}
-
 /// Transform a local clip region to world space (axis-aligned bounding box).
 ///
 /// When the transform includes rotation, the clip becomes the AABB of
 /// the rotated rectangle. This is a conservative approximation that
 /// ensures no clipped content is visible outside the clip region.
 fn transform_clip_to_world(clip: &ClipRegion, transform: &Transform) -> WorldClip {
-    // Transform all 4 corners and compute AABB
-    let corners = [
-        transform.transform_point(clip.rect.x, clip.rect.y),
-        transform.transform_point(clip.rect.x + clip.rect.width, clip.rect.y),
-        transform.transform_point(clip.rect.x, clip.rect.y + clip.rect.height),
-        transform.transform_point(
-            clip.rect.x + clip.rect.width,
-            clip.rect.y + clip.rect.height,
-        ),
-    ];
-
-    // Scale corner radius by transform scale
-    let scale = transform.extract_scale();
-
     WorldClip {
-        rect: aabb_from_points(&corners),
-        corner_radius: clip.corner_radius.scaled(scale),
+        rect: transform.map_rect(clip.rect),
+        // Scale corner radius by transform scale
+        corner_radius: clip.corner_radius.scaled(transform.extract_scale()),
         curvature: clip.curvature,
     }
 }
--- a/src/renderer/paint_context.rs
+++ b/src/renderer/paint_context.rs
@@ -117,15 +117,49 @@ impl<'a> PaintContext<'a> {
     pub fn apply_transform(&mut self, transform: Transform) {
         if !transform.is_identity() {
             self.node.local_transform = self.node.local_transform.then(&transform);
+            self.cull_rect_follows(transform);
         }
     }
 
     /// Apply a transform with origin by composing it with the existing transform.
     pub fn apply_transform_with_pivot(&mut self, transform: Transform, origin: Pivot) {
         if !transform.is_identity() {
-            self.node.local_transform = self.node.local_transform.then(&transform);
             self.node.pivot = origin;
+            self.apply_transform(transform);
         }
+    }
+
+    /// Move the cull rect with the transform just applied.
+    ///
+    /// The rect names the visible region in *this node's* coordinates, and a
+    /// widget that transforms itself has just changed what those coordinates
+    /// mean. Without this the rect goes on describing where the content would
+    /// have been drawn, and everything downstream that narrows children to it
+    /// — the binary-search window and the per-child cull — answers about the
+    /// wrong ones.
+    ///
+    /// The inverse, because the rect travels the other way from the content: a
+    /// subtree lifted 300px upward shows what lies 300px further down. Resolved
+    /// about the node's own pivot with `Transform::about`, the same composition
+    /// `flatten` performs, so a rotation about a corner and one about a centre
+    /// do not agree here and disagree on screen.
+    ///
+    /// A transform with no inverse — a scale of zero — collapses the content to
+    /// a point, so none of it is visible and the rect that says so is the empty
+    /// one. Dropping the rect instead would mean *nothing* is narrowed, and a
+    /// collapsed subtree is a shipped idiom here: `scale(0.0)` is how a menu
+    /// closes, and two hundred rows nobody can see would be painted on every
+    /// frame that dirtied them.
+    fn cull_rect_follows(&mut self, transform: Transform) {
+        let Some(cull) = self.cull_rect else {
+            return;
+        };
+        self.cull_rect = Some(
+            transform
+                .about(self.node.pivot, self.node.bounds)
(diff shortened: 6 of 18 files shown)


MalpenZibo:
## Second quality pass

`/simplify` ran again over everything written after the review rounds — the `own_paint_reach`/`children_outset` split, `gather_reach_upward`, `measure_children`, and the untracked reads, none of which the first pass had seen. Four agents; the deduped set is applied.

**One was a self-inflicted performance regression.** The shrink path put an O(children) fold back on the *per-frame* path. A spring-animated `translate` retargeted downward has its reach decrease on **every frame** of the return leg, not once when it settles — so `measure_children` ran per frame per ancestor, and a staggered list of N rows animating together is O(n²) per frame. Fixing finding 4 from the last round had bought a worse problem.

Two changes close it:
- **Change is detected against the union, not the half being written.** `set_own_paint_reach` compares `paint_overflow` before and after; a widget whose children already reach further has moved nothing an ancestor can see, and the walk is skipped entirely. `grew` now also falls out of the right number — it was being decided from the own half while ancestors consume the union.
- **The shrink path prunes.** A child that was not the widest on either axis cannot have changed a maximum by getting smaller, so it returns without folding. That is the common case: one animating row among many.

**A rename the pass insisted on, three agents independently.** `set_paint_overflow` wrote `own_paint_reach` while `paint_overflow` returned the union with `children_outset` — setter and getter had stopped being inverses while sharing a name, on public API that the book and the widgets skill both instruct external authors to call. It is `set_own_paint_reach` now, named for the half it writes.

**Docs that had stopped being true.** `children_reach`'s doc still said *"one that shrank can wait"* — the shrink path and its test exist to make that false. And `own_paint_reach` carried the stranded paragraph about damage reading it, which is now the union's job. Both were arguing against the code beside them.

**A question answered in the opposite direction to my guess.** I had flagged `Text`/`TextInput` publishing from `layout` as a second way to do one thing. It is not: their reach follows layout-tracked signals and cannot change without a layout, so `refresh_paint_bounds` would recompute a value that never moved. One channel, two schedules — and the trait doc, the book and the skill now say which schedule a new widget belongs on, because mine read as a prohibition the built-in text widgets visibly broke.

Smaller: `Rect::outset_beyond` replaces two copies of the same rect arithmetic; `Rect::ZERO` dropped for the `Rect::default()` the codebase already uses; `measure_children` down to one lookup per child from four, with the loop-invariant parent box hoisted; `gather_reach_upward`'s dead `unchanged` branch collapsed; a vestigial `ordered` flag that was a constant wearing the name of the property; `remeasure_children` now says why layout-only is sound (layout is bottom-up); one `lifted_box` fixture for three tests.

All five mechanisms re-verified by removal after the rewrite, since it touched every one of them:

| removed | result |
| --- | --- |
| untracked reads | `got [Layout]` |
| `refresh_paint_bounds` hook | `left: 0.0, right: 30.0` |
| `gather_reach_upward` | `left: 0.0, right: 70.0` |
| the shrink path's re-measure | `left: 70.0, right: 0.0` |
| `remeasure_children` in publish | `left: 0, right: 1` |

Suite, clippy, goldens on lavapipe, `documentation_references` and `mdbook test` all green. No reference moved beyond the one row already blessed.

**Left open, for the PR body rather than fixed:** during a full layout the upward gather is partly redundant with `remeasure_children`, since each ancestor overwrites what was gathered when its turn comes. The early-out fires immediately when nothing is transformed, so the common case is cheap, but the redundancy is real for a subtree of transformed boxes.

MalpenZibo:
## Fourth review — two blocking findings, both fixed

Both were mine, both found by the reviewer running probes rather than reading, and both are now guarded by a test that fails when the fix is undone.

**1. The damage rect was grown by the *root's* reach, not the widget's.** `surface_relative_bounds_and_root` walks to the surface root, so by the closing line `current` **is** the root — and I had written `paint_overflow(current)` where `main` read the widget's own slot. I introduced it mechanically while repairing an unrelated compile error.

It reintroduced exactly the defect the field's own doc describes: a repaint that "tells the compositor to re-composite a rect that stops short of the shadow, leaving the old one on screen as a fringe". Every elevated card that does not overhang its ancestors damaged short.

Nothing caught it, and the reviewer showed why: the existing test puts its widget directly under the root, so the reach gathers into the root's and both spellings give the same number. `damage_is_grown_by_the_widget_s_own_reach_not_the_root_s` uses a card inset from its parent's edges — with the bug back it reports `80,80 40x20` instead of covering the shadow.

**2. `gather_reach_upward` threaded the wrong quantity, so a shrink stopped one level up.** The guard needs the child's previous `paint_overflow`; I was carrying the parent's old `children_reach`, which is a maximum over *its* children. The two differ wherever a node's children stand further outside it than they themselves reach — which is every scroller over a longer column.

Introduced by the shrink-path prune added in the `/simplify` pass; before that the path always re-measured and was right. The direction was safe — ancestors stayed too wide, so nothing visible was culled — but it falsified the claim that the path stops a reach sticking wide. `a_reach_that_shrinks_releases_every_ancestor` reproduces the reviewer's own probe: with the wrong quantity threaded, the grandparent `stayed at 470`.

## Worth answering, acted on

**A clipping container no longer carries its content's overhang.** This only became live once finding 1 was fixed — the bug was masking it. A 100×100 scroller over 20 rows was reporting **900 px** of paint outside its own box, so it would have damaged and narrowed against a rect nine times its size on every repaint. What a scroller paints is its own box, however far the column inside it runs. `a_clipping_container_does_not_carry_its_content_s_overhang` measures `900 → 0`.

**`measure_children` and the grow path now agree** about a child with no `cached_size`: the reach folds either way, and only the outset needs a box to be measured against. They disagreed silently before.

**The `sorted_axis` doc line** that the `/simplify` pass dropped is back — the early-out it describes is still there.

## Worth answering, not acted on

- **`apply_transform`'s pivot convention.** `cull_rect_follows` resolves each applied transform about the pivot as it stands; `flatten` resolves the accumulated transform about the final one. Conjugation distributes over composition, so the two agree for any number of calls with a single pivot, and `Container` calls it once. They diverge only if something calls `apply_transform` twice with a `set_pivot` between, which nothing does.
- **The issue asked for a scroll event; the tests do two paint frames.** The two-frame form is the sharper half — it is what makes the window and the per-child cull take turns, which a single frame cannot show. A scroll moves the cull rect under the content, which `tests/scroll_virtualization.rs` already covers on the untransformed path. Worth adding, not blocking.
- **Commit 1's message overstates its reach.** It says `Transform::about` replaced four transcriptions; it converts one, and the other three arrive with their callers in commit 2. Both helpers are `pub`, so nothing warns. Accurate to say commit 1 introduces the spelling and commit 2 adopts it.

## Correction to this PR's own body

The snapshot's rows carry `i` pixels of translate, not one pixel each — `.translate((0.0, i as f32))`. That is where the widest reach of 19 comes from, and 19 on a 33 px pitch is the whole justification for the one added row.

All mechanism-removal claims re-verified after these changes.

MalpenZibo:
## The mutation job's survivors, closed

The last run reported **41 missed of 150** — and unlike a feature-gated module, all of it compiled and reachable. The clusters were the two mechanisms most likely to be wrong, so they are now covered rather than carried as a follow-up.

**The swept-circle rotation bound was watched by nothing.** `reach_tests` reached `outset_of` directly and never the branch in `max_transform_reach` that computes it, so ~18 mutants in the corner-distance arithmetic survived. `the_rotation_bound_covers_every_angle_for_the_pivot_it_is_given` brute-forces all 360 angles for four pivots and asserts the bound covers the worst — **and is tight**. The tightness half is what does the work: a bound asserted only to be *large enough* is satisfied by any arithmetic that errs upward, which leaves the computation unwatched and pays for reach it cannot use on every rotation. `Pivot::BOTTOM_RIGHT` is in the list deliberately — it is the one whose circle reaches furthest past the far edges rather than the near ones.

**The prune guard in `gather_reach_upward`.** Six survivors on the comparison the third review singled out as the one that could cull something visible. My first attempt at a test could not distinguish them: both halves of the guard sat on their boundary together, so the second rescued the first and a wrong comparison in either was invisible. The rewritten pair picks children whose halves disagree — one stands furthest outside the box, the other reaches furthest:

- `the_widest_child_shrinking_is_not_pruned_away` kills `<` → `<=`, `<` → `==`, and `&&` → `||`.
- `a_child_that_leads_on_outset_alone_is_not_pruned_away` kills the same on the second half — a child can hold up either maximum, and pruning on the reach alone leaves the outset stale.
- `a_child_that_was_not_the_widest_leaves_the_parent_alone` pins the prune actually pruning, which is what keeps a spring's return leg off an O(children) walk per frame.

**The window's near-side widening.** `near - children_reach` → `near + children_reach` survived: nothing asserted the search widens on the near side at all, only the far one, because everything below the fold is a far-end case. `the_window_widens_on_the_near_side_as_well` puts a row above the visible span whose shadow hangs down into it. Its reach is an **elevation**, not a transform, which is the point — what the window widens by is how far a child paints outside itself, whatever put it there.

Each of these was verified by applying the mutant and watching the named test fail, not by assuming.

## Two survivors left, and why they stay

`far * 2.0` → `far / 2.0`, on the swept rect's width and on its height. The swept rect is a square centred on the pivot, so for a corner pivot both far-side terms equal `far` and halving one leaves the other holding the maximum. I could not construct a box and pivot where either is the unique maximum, which makes them look equivalent rather than unwatched. Recording that rather than contorting a test until it goes green.

Two more are equivalent by inspection: `was > had_reach` in the prune (it simply always re-measures — correct, slower) and `after > before` → `>=`/`==` in `set_own_paint_reach` (the equal case has already returned).

MalpenZibo:
## Where the mutation job landed

**41 → 19 → 17 → 12 missed, 137 caught.** Every remaining survivor is equivalent or fails in the safe direction, checked one at a time rather than assumed:

- `>` → `>=` in `set_own_paint_reach` and `remeasure_children` — the guard above has already established the two values differ, so it is the same predicate.
- `<` → `>` in the prune guard, `>` → `==`/`<` in `remeasure_children` — these can only make `grew` *false* when it should be true, which takes the re-measure path: correct, merely more work. No mutant can make it wrongly true, which is the direction that would drop a widget.
- `Moves::any` → `true` — the early-out stops firing, and the arithmetic below then computes identity transforms and returns the same number.
- `Widget::refresh_paint_bounds` → `()` — the default body is already a no-op.
- `far * 2.0` → `far / 2.0` on the swept rect's width or height — the rect is a square centred on the pivot, so for a corner pivot both far-side terms equal `far` and halving one leaves the other holding the maximum. No counterexample found.

Closed since the previous run: `remeasure_children`'s upward gather (added a round earlier with no test at all, which is the pattern that has caught something every round) and the spring overshoot — a `BOUNCY` translate to −100 passes −100 on the way, so a reach of exactly 100 culls the widget at the top of its own bounce.

## A correction

An earlier comment reported `>` → `<` as killed. It was not. The script that applied it replaced *both* occurrences of that expression, and the failures came from the one in `set_own_paint_reach` rather than the one under test in `remeasure_children`. The conclusion happened to be right — the mutant is equivalent — but the evidence for it was not, and the two are different things.

## What is not reviewed

The delta after the fifth review — the `clips_children` flag on `Slot`, `remeasure_children` gathering upward, and `lengths.overflow` replacing an untracked read — has not been through a review pass of its own. One flag, one early-out and one substituted expression, each with a test verified by deleting the mechanism and watching it fail.

Issue #330, opened 2026-09-04
A signal created during registration belongs to the surface, not to the widget


MalpenZibo:
## Observed

`register_children` runs *outside* the owner the widget was built in, so a
signal created there is filed under the surface and outlives the widget.

`ChildrenSource` wraps the widget *factory* in `with_owner` and hands back an
`OwnedWidget` (`src/widgets/children.rs:203-211`); `register_children` is called
afterwards, and `OwnedWidget::register_children` (`children.rs:507`) forwards
without re-entering the owner. `register_signal` (`src/reactive/owner.rs:377`)
files under `current_owner()`, which by then is the surface's.
`teardown_widget_subtree` (`src/jobs.rs:369`) clears subscribers and unregisters
the widget; it does not free the signal slot.

Found while adding `Container::enabled` (#242). `Container::fold_enabled`
(`src/widgets/container/mod.rs`) builds a `create_derived` at registration —
the only `create_derived` in the widget layer, so this is a gap nothing had
stood in before rather than a regression.

## Who is worse off

An application with a keyed list whose rows declare `enabled` inside a control
that also declares it. Every row created leaks one derived slot and its
subscriber entry, for the surface's lifetime. The captured environment is two
`Copy` signals — 24 bytes — so nothing large is held; the growth is the slot and
the map entry, and it is unbounded over a long session with churn.

Nothing else in the widget layer creates a signal during registration today, so
the bound is exactly that shape. It is narrow, and it is also the kind of thing
that quietly stops being narrow: the fold is the natural place for any future
"resolve this against my ancestors once" answer, and each one would land here.

## Expected

A signal created during `register_children` is disposed when the widget that
caused it is, exactly as one created in the build closure is.

## Acceptance

A test that builds a keyed list whose rows declare `enabled` beneath a
container that declares it, churns the keys N times, and asserts the reactive
runtime's live-signal count returns to its starting value — rather than growing
by one per row created. It fails today.

## Where it is

- `src/widgets/children.rs:203-211` and `:507` — where registration happens
  outside the owner
- `src/reactive/owner.rs:220` — `with_owner` creates a *new* owner; there is no
  `run_with_owner(existing)` to re-enter one, which is the missing piece
- `src/widgets/container/mod.rs` — `fold_enabled`, the one caller that notices

## Note

This is a change to when a cross-cutting mechanism runs — either
`OwnedWidget::register_children` re-enters its owner, which needs a way to
[...]

PR #331, opened 2026-09-04
A subtree can be told not to take input, and a field can be told not to take edits

### LINKED ISSUE #211: A text input cannot be told to stop accepting keys
> As an application author, I want to declare that an input is not accepting
> keys, so that a field whose answer is already in flight stops taking
> characters that can never be sent.

### Where it came from

allock's `Auth::submit` sets `busy` and refuses a second Enter, and the doc on
`AuthState::busy` says "the field is disabled until the backend answers".
Nothing disables it. PAM takes a second or two over a wrong password *on
purpose*, and everything typed in that second is accepted, masked, drawn, and
then discarded by `input_w.set(String::new())` when the answer arrives.

### Confirmed

`TextInput`'s public setters are `password`, `mask_char`, `widget_ref`,
`placeholder`, `no_caret`, `autofocus`, `on_change`, `on_submit`. There is no
way to say "not now".

`WidgetRef::blur()` is the nearest thing and is a different thing: it drops the
focus, and with it the `when_focused` layer that on a multi-monitor lock screen
is the only thing saying which screen the keys are going to. It also has to be
handed back afterwards, against an `autofocus` with its own opinion about when.

### One thing the story understates

The *look* is already expressible: `StateWhen::When(Signal<bool>)` exists, so
`when(busy, |s| s.darker(0.2))` styles a disabled field today. What is missing
is only the **behaviour** — nothing stops the keys.

That is worth deciding on purpose rather than by default. Two shapes:

- `enabled(signal)` on `TextInput` alone, and applications keep expressing the
  look through `when(..)`. Smallest change, and the look stays the app's.
- `enabled(signal)` plus a `StateWhen::Disabled` that follows from it, so the
  look is declared once beside the behaviour. More consistent with how `Hovered`,
  `Pressed` and `Focused` already pair a state with its styling — and it makes
  `Disabled` mean the same thing everywhere, which `When(busy)` cannot promise
  across two widgets in different files.

The second is the one that matches the rest of the state layer API. It also
raises the question of what `Disabled` means for a container that only has an
`on_click` — worth answering in the same pass rather than later.

### Done when

`text_input(sig).enabled(signal)` refuses keys, paste and pointer edits
reactively, while keeping the focus it holds.

### LINKED ISSUE #242: Nothing can be declared disabled, so nothing below it can be either
Carried in `docs/REFACTORING.md` under "Still open" until that document was
deleted. Verified still open: `disabled` appears nowhere in `src/` except in
messages about cargo features.

## Observed

There is no way to declare that a widget, or a subtree, is not accepting input.
An application that wants a form to stop responding while a request is in
flight threads a signal into every handler by hand and checks it there.

The *look* is already expressible — `StateWhen::When(Signal<bool>)` exists, so
`when(busy, |s| s.darker(0.2))` greys a field today. What is missing is only the
behaviour.

## What the earlier note got wrong

It said `disabled` propagation is event dispatch rather than style resolution,
and that putting it in dispatch "touches how every event is routed". Reading the
code, the first half is right and the second is not.

`Container::event` already opens with the gate this needs, for a different
property:

```rust
// src/widgets/container/mod.rs:1515
fn event(&mut self, tree: &mut Tree, id: WidgetId, event: &Event) -> EventResponse {
    if !self.visible.get_or(true) {
        return EventResponse::Ignored;
    }
```

Dispatch is one recursive walk — `dispatch_events` (`src/lib.rs:625`) hands the
event to the root and every `Container::event` passes it to its children. Keys
take the same path as the pointer; there is no separate focus route. So a
subtree gate is a second condition on an early-out that already exists, and the
routing is untouched.

## Where the decision actually is

Not in dispatch. In how a *style* asks "am I disabled?", because that question
has to be answerable **with no tree in hand**:

```rust
// src/widgets/container/mod.rs:359, on StateWhen::Focused
// The path, rather than a walk of this container's descendants: the same
// question has to be answerable from a `create_derived` closure, which has no
// tree, and that is where a container resolves the text colour it publishes
// below it.
```

The three existing states each answer it differently, and both mechanisms are
worth naming because a fourth state has to pick one:

- **`Hovered`, `Pressed`** — `Control` holds an `RwSignal<InteractionFlags>`;
  reading it is tree-free and is what subscribes the reader.
- **`Focused`** — `request_focus(tree, id)` resolves the ancestor path *once*,
  where the tree is in hand, and stores it. `focus_path()` then reads a global
  signal.

`disabled` is like `Focused` in shape: the answer depends on ancestry, so it has
to be resolved where the tree is and read where it is not. **Context is not an
option** — `src/reactive/context.rs` is app-global with one value per type, so
it cannot carry a value scoped to a subtree.

The mechanism this implies: `Control`
gains an `enabled: Signal<bool>` built at `register_children` time — where
`Control::new(id, ix.flags)` is already built and the tree is already in hand —
as a derived signal over this container's own declaration **and** its parent
control's. Effective-disabled is then one signal read, tree-free, reactive, and
`nearest_control` keeps working unchanged.

## The decision

**Shape A, agreed with the maintainer: one concept, styleable, propagating.**

- `Container::enabled(impl IntoSignal<bool, M>)` gates the subtree.
- `TextInput::enabled(impl IntoSignal<bool, M>)` is the same setter with a
  smaller scope, which is #211.
- `StateWhen::Disabled` follows from it, so the look is declared beside the
  behaviour and `Disabled` means the same thing everywhere — which
  `When(busy)` cannot promise across two widgets in different files.
- **Disabling propagates downward and a descendant cannot undo it.** A
  container declaring `enabled(true)` inside one declaring `enabled(false)` is
  disabled.

It pairs a state with its styling exactly as `Hovered`, `Pressed` and `Focused`
already do, which is the shape the rest of this API has.

Two alternatives were weighed and are recorded so they are not re-opened by
accident. **B** was the web platform's split — a `Container::inert(signal)` that
gates the subtree and is deliberately not styleable, beside a styleable
`TextInput::enabled`. It was rejected because the reason the web keeps `inert`
unstyleable is that `inert` also strips a subtree from the accessibility tree,
and guido has no accessibility tree for it to strip — so the argument does not
transfer, and one concept is fewer than two. **C** was #211 alone, with no
subtree gate; it is the smaller half of A rather than a different answer, and
leaves this issue open.

## Acceptance

Every one of these is a test in `src/widgets/container/characterization.rs`
unless said otherwise. Each fails today because the setter does not exist.

- **A click does not reach a handler below a disabled container.** `on_click` on
  a child, `enabled(false)` on the parent, a `MouseDown`/`MouseUp` pair inside
  the child's bounds: the callback does not fire and the root's response is
  `Ignored`.
- **Nor does a key.** A `TextInput` under a disabled container, focused, sent a
  `KeyDown`: its signal is unchanged. In `tests/text_input_editing.rs`, which
  already drives that widget.
- **Disabling propagates and cannot be undone from below.** A container
  declaring `enabled(true)` inside one declaring `enabled(false)` is disabled.
  This is the rule Qt, GTK and SwiftUI all settled on; asserting it is what
  stops it being re-litigated per widget.
- **It is reactive, and costs no layout.** Flipping the signal back lets the next
  click through, and the write queues no `JobType::Layout` — asserted with
  `h.jobs_from(..)`, as the existing reach tests do.
- **A container disabled while hovered does not stay hovered.** The `visible`
  early-out returns before `track_pointer` runs, so a naive copy of it strands
  `InteractionFlags::HOVERED` set and `when_hovered` sticks for ever. After
  disabling, `Control::is_hovered()` is false.
- **Focus is kept, not dropped.** #211 is explicit that a field which stops
  taking keys must keep the focus it holds — on a multi-monitor lock screen the
  `when_focused` layer is the only thing saying which screen the keys were going
  to. `focus_path()` is unchanged across the flip.
- **A descendant can style itself from it.** A `Text` inside a
  disabled container resolves `StateWhen::Disabled` as active, through its
  control, with no tree — the `create_derived` constraint above.
- **A `compile_fail` doctest**: `when_hovered(|s| s.enabled(false))` does not
  compile. Enabling is a declaration, not a state override, the same way a
  timing is.
- Book chapter and `tests/documentation_references.rs` green.

## Non-goals

- **Accessibility.** guido has no accessibility tree, so nothing here is about
  removing a subtree from one. That is what `inert` is for on the web and it is
  not what this is.
- **Focus navigation.** There is no tab order to skip: `Key::Tab` is delivered
  as a key and nothing moves focus with it (`src/platform/input.rs:859`).
- **Painting.** A disabled subtree is drawn exactly as declared. Greying it out
  is the application's, through the state layer.
- **`enabled` on every widget type.** `Container` and `TextInput`; `Text` and
  `Image` take no input to refuse.
- **Scroll.** Whether a disabled subtree still scrolls is deliberately left to
  the implementation to answer with a test, and is a sentence in the pull
  request either way.

## Where it lives

- `src/widgets/container/mod.rs` — the `visible` early-out at `:1515` that this
  joins, `is_control` at `:1078`, and `register_children` at `:1331` where
  `Control` is built.
- `src/widgets/control.rs` — `Control`, which would carry the effective answer.
- `src/widgets/state_layer.rs` — `StateWhen` at `:75` and `is_active`.
- `src/widgets/text_input.rs:1344` — `TextInput::event`, for #211's half.
- `src/lib.rs:625` — `dispatch_events`, for context; unchanged by this.

## Prior art

The web platform splits this in two, deliberately:

- [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled)
  is a *form-control* concept. It is styleable through `:disabled`, and on
  `<fieldset>` it propagates to every descendant control except those inside the
  `<legend>`.
- [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert)
  is a *subtree* concept, on any element. It blocks all pointer and click
  events, removes the subtree from the tab order, and hides it from the
  accessibility tree. There is deliberately **no `:inert` pseudo-class** — it is
  not meant to be styled, because what it primarily does is an accessibility
  act.

Every desktop toolkit chose one concept instead, and all of them propagate it
downward and refuse to let a descendant re-enable itself:

- **Qt** — `QWidget::setEnabled(false)` implicitly disables all children, and a
  child cannot override it. Qt documented this explicitly after it came up
  repeatedly ([qtbase
  c0e04ba](https://git.sailfishos.org/mer-core/qtbase/commit/c0e04bad701211613269a5014c593c1d4d25a619)).
- **GTK** — `gtk_widget_set_sensitive()`, with the same descendant behaviour.
- **SwiftUI** — [`.disabled(_:)`](https://developer.apple.com/documentation/swiftui/environmentvalues/isenabled)
  writes the `isEnabled` environment value down the hierarchy, and a nested
  `.disabled(false)` does *not* re-enable; the only escape is writing the raw
  environment value, which is a lower-level act.
- **Flutter** — no inherited property: `AbsorbPointer`/`IgnorePointer` wrap a
  subtree, and Material widgets take `onPressed: null` to mean disabled.
- **iced, Compose** — per-widget only (`on_press(Option<Message>)`, an `enabled`
  parameter). No subtree concept at all.

So the propagation rule agreed above is not a guess: it is what four of the five
settled on, and the one that differs — the web's `inert` — differs for a reason
guido does not have.

### Related

#211 is the narrow half — one text input that can be told to stop accepting
keys. Under the agreed shape it is the same setter with a smaller scope, so it
ships as part of this change rather than before it, and closes with it.

MalpenZibo:
## What changed

A subtree can be declared not to take input — `Container::enabled(signal)` — and
it propagates: a descendant declaring `enabled(true)` inside one declaring
`enabled(false)` is disabled. `StateWhen::Disabled` and `Stateful::when_disabled`
style it, so the look is declared beside the behaviour rather than threaded into
every handler by hand. Separately, `TextInput::readonly(signal)` refuses every
edit while keeping the focus.

**This deviates from the shape #242 agreed, deliberately.** #242 specified one
hybrid concept: `enabled` refuses input but *keeps* the keyboard focus, and
`TextInput::enabled` as the same setter with a smaller scope. The first review
found that promise was already false — `dispatch_events` (`src/lib.rs:643`)
releases the focus on any `MouseDown` the root does not return `Handled` for, so
a press on a disabled panel dropped the very ring the feature existed to keep. I
reproduced it through `Headless`: enabled, a press on a focus box's padding keeps
the focus; disabled, the same press drops it.

Researching that turned up the real problem, which #242's prior art had missed by
looking at the wrong axis. It weighed `disabled` against `inert` — one control
versus a subtree — and never against `readonly`, which is where the focus
question actually lives. Every toolkit splits these two, and splits them
*because* of the focus difference: `readonly`/`disabled` on the web,
`setReadOnly`/`setEnabled` in Qt, `readOnly`/`enabled` in Flutter,
`editable`/`sensitive` in GTK. Guido's hybrid would have been the only one of its
kind, and the requirement that produced it — allock's field must stop taking keys
while keeping the ring that names the monitor — is a description of `readonly`,
not of `disabled`.

So the shape agreed in this session, and what is here:

- `Container::enabled` is the propagating disable, and gives up the keyboard as
  well as the pointer: a disabled unit resolves no `when_hovered`, no
  `when_pressed` and no `when_focused`. That is Qt, GTK, SwiftUI and
  `<fieldset disabled>`, and it makes the press question moot — there is no ring
  left to protect.
- `TextInput::enabled` became `TextInput::readonly`, which is what closes #211.

`#242` still describes the old shape; it is superseded by this, not contradicted.

**Scroll**, which #242 left to the implementation: a disabled subtree does not
scroll. It follows from where the gate sits rather than from a separate decision,
and `a_disabled_scroller_does_not_scroll` says so.

Closes #242. Closes #211.

## What proves it

Every acceptance bullet in #242 has a test that fails without the change.

In `src/widgets/container/characterization.rs`:
`a_click_does_not_reach_a_handler_below_a_disabled_container`,
`a_descendant_declaring_itself_enabled_stays_disabled`,
`re_enabling_lets_the_next_click_through_and_costs_no_layout`,
`a_container_disabled_while_hovered_is_no_longer_hovered`,
`a_text_below_a_disabled_container_styles_itself_from_it`,
`a_disabled_scroller_does_not_scroll`.

In `tests/text_input_editing.rs`:
`a_field_below_a_disabled_container_does_not_hear_the_keyboard`,
`a_disabled_container_stops_claiming_the_focus_below_it`,
`a_read_only_field_refuses_every_edit`,
`a_read_only_field_refuses_a_middle_click_paste`,
`a_read_only_field_keeps_the_focus_and_the_ring`,
`a_read_only_field_still_moves_its_caret`,
`a_read_only_field_still_submits`.

Plus the `compile_fail` doctest that `when_hovered(|s| s.enabled(false))` does
not compile, the three spellings in `tests/signal_conversions.rs`, and
`mdbook test` over the new chapter.

No golden or snapshot moved, and none should have: nothing in the renderer,
geometry or `src/platform/` changed. The style resolution is watched at the
render-tree level by `a_text_below_a_disabled_container_styles_itself_from_it`,
which reads a painted colour.

## What the review found

The `reviewer` subagent ran twice, because the design changed between them.

**First pass: one blocking finding** — the focus promise above. It is what sent
this back to the maintainer and produced the readonly split. Also raised, and
acted on: `TextInput`'s own state arm had no test (it no longer exists — the arm
collapsed into the generic one when `enabled` left the widget).

**Second pass, over the reworked change: one blocking finding.** A middle click
pastes the primary selection straight into `insert_text`, bypassing the guard
that sat at `handle_key` — on a lock screen, into the password PAM is answering.
Fixed by moving the refusal to the four functions that write the text, so a
future fifth entrance is covered by construction;
`a_read_only_field_refuses_a_middle_click_paste` fails without it. Its two
"worth answering" findings are both acted on: `Enter`, `Ctrl+X` and `Ctrl+Y` are
now exercised (`a_read_only_field_still_submits` and the extended refusal test).

Its remaining notes: `tests/signal_conversions.rs` carries the spellings for both
setters in commit 4 rather than in the commits that introduced them, so neither
of those two is revertible entirely on its own. The spellings are present and
pass, which is what that rule is for.

Before either review, `/simplify` applied four passes. The one that mattered
structurally: `Control::is_active` is now the single statement of what each
`StateWhen` means, where the container, `Text` and `TextInput` each spelled it
out separately before — that is commit 1, and it stands on its own.

## What was checked by hand

Nothing on a compositor. `examples/disabled_form.rs` exists and builds — it puts
a disabled panel and a read-only field side by side, which is the only way the
difference is obvious — but I did not run it, so the claims in its header are
asserted by the tests above rather than by my having looked at a surface. Worth a
minute on niri before merge.

One thing was checked below the harness rather than by eye: the focus behaviour
that sent this back was reproduced through `guido::testing::Headless`, driving
the real loop and `dispatch_events`, because `Harness::send` calls
`widget.event` directly and cannot see the dispatcher's rule. That probe was
thrown away once the design changed and the behaviour it measured no longer
exists.

## Left undone

**A descendant of a newly-disabled subtree is never told the pointer left.**
`pointer_left` runs on the gating container only, and the gate returns before the
children are asked — so a `TextInput` that was hovered keeps its I-beam over a
field that takes nothing. `visible(false)` strands identically today, which is
why this is not filed against this branch: it belongs to the shared early-out,
and a fix there wants an acceptance criterion covering both. Named here so the
next reader of that gate does not re-find it.

**#330** — a signal created during `register_children` is filed under the surface
rather than the widget, so a keyed list whose rows declare `enabled` beneath a
control that also does leaks one slot per row. Found at the definition,
documented there, and filed rather than left in this body. Closing it means
running registration under the widget's own owner, which is a cross-cutting
change and the maintainer's call.

**A disabled field keeps blinking its caret**, since it keeps the focus it holds
in guido's internal sense even though no ring resolves. `.caret(move || !busy.get())`
expresses it, which is why the caret was deliberately left out of `readonly`
rather than bundled into it. No issue: nobody is worse off who cannot say so in
one setter.

https://claude.ai/code/session_016cdRGJettCQPPACDkS1CtM

### DIFF
--- a/.claude/skills/widgets/SKILL.md
+++ b/.claude/skills/widgets/SKILL.md
@@ -12,8 +12,8 @@ declarations do not.**
 
 Reactive: `background`, `gradient`, `backdrop_blur`, `overflow`, `corners`,
 `border`, `translate`, `rotate`, `scale`, `pivot`, `width`, `height`,
-`padding`, `visible`, `shadow` — and beyond `Container`, `Text`'s `wrap`,
-`Image`'s `content_fit`, `TextInput`'s `password`, `mask_char` and `caret`,
+`padding`, `visible`, `enabled`, `shadow` — and beyond `Container`, `Text`'s `wrap`,
+`Image`'s `content_fit`, `TextInput`'s `password`, `mask_char`, `caret` and `readonly`,
 `RippleConfig`'s colour, and the direction a `Flex` is built with.
 
 Structural: `layout`, `child`, `children`, `control`, the axis a `Scroll` is
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -33,6 +33,7 @@
 - [Interactivity](interactivity/README.md)
     - [State Layer API](interactivity/state-layer.md)
     - [Hover & Pressed States](interactivity/states.md)
+    - [Disabling a Subtree](interactivity/disabled.md)
     - [Ripple Effects](interactivity/ripples.md)
     - [Event Handling](interactivity/events.md)
 
--- a/book/src/interactivity/README.md
+++ b/book/src/interactivity/README.md
@@ -30,6 +30,7 @@ The framework handles:
 
 - [State Layer API](state-layer.md) - Overview of the state layer system
 - [Hover & Pressed States](states.md) - Define visual overrides per state
+- [Disabling a Subtree](disabled.md) - Stop a whole subtree taking input, and style it
 - [Ripple Effects](ripples.md) - Material Design-style touch feedback
 - [Event Handling](events.md) - Click, hover, and scroll events
 
--- a/book/src/interactivity/disabled.md
+++ b/book/src/interactivity/disabled.md
@@ -1,0 +1,102 @@
+# Disabling a Subtree
+
+A form that stops responding while a request is in flight is one declaration,
+not a check inside every handler:
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+# let busy = create_signal(false);
+# let submit = || {};
+container()
+    .enabled(move || !busy.get())
+    .child(container().on_click(submit).child(text("Send")))
+# ;
+# }
+```
+
+`enabled(false)` stops every event at that container. Nothing below it — a
+click, a key, a scroll, a nested button's handler — is reached, and nothing
+below it has to know why.
+
+## It propagates, and cannot be undone from below
+
+A container declaring `enabled(true)` inside one declaring `enabled(false)` is
+**disabled**. There is no way for a descendant to re-enable itself, which is
+what lets a whole form be switched off in one place and is the rule Qt, GTK and
+SwiftUI all settled on.
+
+## The look is declared beside the behaviour
+
+A disabled subtree is painted exactly as declared — guido does not pick a grey
+for you. `when_disabled` is how you pick one, and it reads the same declaration
+that stopped the events, so the two cannot come apart:
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+# let busy = create_signal(false);
+container()
+    .enabled(move || !busy.get())
+    .child(
+        text("Send")
+            .color(Color::WHITE)
+            .when_disabled(|s| s.color(Color::rgb(0.5, 0.5, 0.5))),
+    )
+# ;
+# }
+```
+
+The text asks its **control** — the nearest enclosing interaction unit, which
+declaring `enabled` makes the container into. That is why the label greys with
+the form around it without being handed the signal itself.
+
+A condition you already hold reads the same at one call site:
+`state(busy, |s| s.darker(0.2))`. What it cannot do is promise that two widgets
+in two files mean the same thing by it. `Disabled` is the control's answer, with
+its ancestors folded in, so it means one thing everywhere.
+
+## What a disabled subtree gives up, and what it keeps
+
+**It gives up the pointer and the keyboard.** A disabled control is neither
+hovered nor pressed nor focused, so `when_hovered`, `when_pressed` and
+`when_focused` all switch off with it — no stale highlight, and no focus ring on
+something that is not taking keys. Qt, GTK, SwiftUI and `<fieldset disabled>`
+all behave this way.
+
+**It is still laid out and still painted.** Disabling changes what an event
+does, not where anything sits — so it costs no layout pass.
+
+## Not the same as read-only
+
+There is a second thing you may want, and it is not this one: a field that stops
+accepting *edits* but goes on saying the keyboard is aimed at it. A lock screen
+that has sent the password to PAM wants exactly that — the typed characters must
+go nowhere, but on a multi-monitor setup the `when_focused` ring is the only
+thing saying which screen you are typing at, and disabling would take it away.
+
+That is `readonly`, and it is a different setter for a different job:
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+# let value = create_signal(String::new());
+# let busy = create_signal(false);
+text_input(value).readonly(busy).caret(move || !busy.get())
+# ;
+# }
+```
+
+It refuses every edit — typing, backspace, delete, paste, cut, undo — and
+nothing else. The caret still moves, a selection can still be made and copied,
+`Enter` still reaches `on_submit`, and the focus stays put. The caret is not
+part of it: a field that should stop blinking says so with `caret`, driven by
+the same signal, as above.
+
+Every toolkit keeps these two apart, and for this reason: `readonly` beside
+`disabled` on the web, `setReadOnly` beside `setEnabled` in Qt, `readOnly`
+beside `enabled` in Flutter. Reach for `enabled` when the thing is not
+available, and for `readonly` when it is yours but frozen.
--- a/book/src/interactivity/state-layer.md
+++ b/book/src/interactivity/state-layer.md
@@ -7,6 +7,8 @@ The state layer system provides declarative style overrides based on widget inte
 State layers let containers define how they should look when:
 - **Hovered** - Mouse cursor is over the widget
 - **Pressed** - Mouse button is held down on the widget
+- **Focused** - The keyboard focus is inside the widget
+- **Disabled** - The widget, or something above it, was declared `enabled(false)` — see [Disabling a Subtree](disabled.md)
 
 Changes are defined declaratively, and the framework handles state transitions, animations, and rendering.
 
--- a/docs/STATE_LAYER.md
+++ b/docs/STATE_LAYER.md
@@ -8,6 +8,7 @@ State layers allow containers to define how they should look when:
 - **Hovered**: Mouse cursor is over the widget
 - **Pressed**: Mouse button is held down on the widget
 - **Focused**: Any child widget has keyboard focus (e.g., text input)
+- **Disabled**: the control, or something above it, was declared `enabled(false)`
 - **A condition the app owns**: `state(condition, |s| ...)` — the last submit failed, this row is selected
 
 Style changes are defined declaratively using builder methods, and the framework handles all state transitions, animations, and rendering automatically.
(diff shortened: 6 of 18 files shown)


MalpenZibo:
Mutation job, after the last commit: **68 mutants, 45 caught, 6 missed, 17 unviable** — up from 38 caught / 13 missed on the first run. The two clusters it found in new code are dead: `mutates` can no longer return a constant unnoticed (the refusal test asserts the *response* now, not just the text), and `TextInput::is_state_active` can no longer be replaced by one (nothing had ever declared a state layer on a field).

The six that survive, so they are not re-found:

- `Container::is_control -> true` — a pre-existing gap on a line this branch touched.
- `mutates`' `ctrl` guard replaced with `true` — the refusal loop only sends a char that is in the ctrl list, so the plain-char path is not separated from it.
- `TextInput::event`'s `MouseLeave` arm deleted — the field's hover clearing on leave is untested, and was before this branch.
- three operator flips in `pointer_left`'s repaint condition — it decides only whether an `Animation` job is requested, and the code is moved rather than written here.

None is a defect: they are coverage, and there is no failing test to write today for any of them. Recorded here rather than as issues, per the bar in AGENTS.md.

Issue #333, opened 2026-09-04
A text's font size is not resolved through the door, so a bad one poisons its animation for good


MalpenZibo:
## Observed

`Text` and `TextInput` resolve their font size with a plain read, so a
non-finite one reaches the animation and stays there.

`src/widgets/text.rs:204` and `src/widgets/text_input.rs:663`:

```rust
style.font_size.get_or(14.0)
```

`refresh` caches the value, `retarget_text_anims` hands it to
`TextAnims::retarget`, and that calls `AnimationState::animate_to`. The early-out
is `new_target == self.target`, which NaN fails against itself, so a segment
begins with `current` stored as `start` — and `lerp` from a NaN start is NaN at
every `t` thereafter. A later, perfectly finite font size then animates from
nothing to nothing for the life of the process.

Reachable with one line of ordinary application code:

```rust
text("x").font_size((move || width.get() / columns.get()).transition(200.0))
```

## Who is worse off

Anyone whose font size is computed rather than constant — a label sized to fit
its column, a lock screen scaling to the output. One frame with a zero divisor
and the text keeps its size for the rest of the session, silently: the picture
is *correct*, it is simply not the one the signal asks for, which is the hardest
failure in this library to find by looking.

The frame loop is not pinned here, unlike #228's transform case: `retarget` runs
from layout rather than from every paint, so the transition completes and
`is_animating` goes false. It is the permanent poisoning, not the battery.

## Why it is not closed by #228

#228 coerced every declared value **where it is resolved**, which is CSS's rule
and is why nothing downstream needed a guard. The cost of that placement is that
it covers the resolvers that were converted, and `Container`'s were.
`Text` and `TextInput` have their own, and `TextStyle` has no state-layer
resolution at all — `inherit_from` only `or`s the options together — so there is
no `resolve_state_value` to hang the door on.

The previous attempt (#225) guarded `animate_to`/`set_immediate` instead, which
would have covered every widget for free. It was backed out because guarding
consumers made each guard change what the next one saw. Closing this the same
way as #228 means giving the text widgets a resolver to put a door in.

## Expected

A text whose font size goes non-finite keeps the size it had, says so once, and
follows the signal again when it recovers — the same contract #228 established
for a container's declared properties.

## Acceptance

In `tests/` or beside the widget: a `Text` with `font_size` driven by a signal,
set to `f32::NAN` for a frame and then to a finite size. The painted glyph size
follows the finite value. It fails today — the animation is poisoned and never
recovers.

[...]

PR #334, opened 2026-09-04
A value nothing can compute is read as if it had not been declared

### LINKED ISSUE #228: A signal that resolves to NaN reaches the matrix, the animation and the damage rect
A signal that computes a value can produce a NaN or an infinity — a division by
a measured zero is all it takes — and nothing in the library refuses one. The
value goes wherever the finite one would have gone.

```rust
container().rotate(move || (a.get() / b.get()).to_degrees())   // b == 0.0
```

### Where it ends up

**In the matrix.** `sin_cos(NaN)` is `(NaN, NaN)`, so the whole affine is NaN.
`is_identity()` compares equal to nothing and answers false, so it is applied.
`inverse()`'s `det.abs() < 1e-10` is false for a NaN, so hit testing returns
NaN coordinates and every `contains` answers false. `world_aabb`'s
`fold(f32::INFINITY, f32::min)` ignores NaN and returns the initial value, so
the rect comes out `-inf` wide — and that rect sizes a damage region.

**In the animation.** Worse, because it persists. `AnimationState::set_immediate`
is the seeding store and writes `current`, `start` and `target` together, and
`advance` computes `lerp(start, target, t)` — NaN at every `t` once `start` is,
with only the spring branch assigning the target outright at the end. So a
later, perfectly finite target animates from NaN to NaN for the life of the
process.

**And it never settles.** `animate_to`'s early-out is `new_target ==
self.target`, which NaN fails against itself, so every pass reads a new target,
begins a segment and asks for another Animation job. The surface never goes
idle: the frame callback stays armed at the compositor's rate for a value that
was never going to arrive.

**`Pivot` is the fourth component and reaches `center_at` directly**, with no
guard of its own even if the other three were sanitised.

### What was tried, and why it is an issue rather than a diff

All of the above was written and then backed out of #225 (see its history for
the working code):

- `Animatable::is_reachable`, refusing a target in `animate_to` and a value in
  `set_immediate`
- a matching refusal in `resync_animation_targets`, without which the drift
  check asks for an Animation job on every paint for a target that was refused
- a finiteness guard at the top of `Transform::compose`, and one in
  `Pivot::resolve`
- a rate-limited diagnostic shared between them

It was backed out because each of those pieces went in as a response to a
review finding on a PR about something else, and each one needed its own
correction in the round after:

- refusing in `animate_to` alone left the drift check waking the loop for ever
- refusing in `set_immediate` left the property `is_initial()` for ever, which
  also switches off drift detection
- refusing an enter left the property pinned at the enter value
- the pivot's fallback was computed from the same bounds that produced the bad
  value, so it could be NaN too
- the diagnostic named neither the property nor the widget while sharing one
  process-wide rate limit with every other caller
- `is_reachable`'s overrides landed on the three cheap types and not on the
  four that actually allocate

Every one of those is fixable. Together they are a design that wants doing in
one piece, not accreting one review round at a time.

### What it needs deciding

- **Where the door is.** One guard at the boundary where a signal's value
  enters the widget (so every property is covered once), or one per consumer
  (so each says something useful about what it did)?
- **What a refusal leaves behind.** The property has to stay usable: seeded,
  not drifting, not pinned, and able to recover when the signal comes back to
  its senses. That is the part every attempt got wrong in a different place.
- **What it says.** `compose` logged; the animation guards did not. A property
  that silently stops following its signal is the hardest failure in this
  library to diagnose, and the diagnostic should name the property and the
  widget rather than the module it happened to be called from.
- **Whether infinity is the same case as NaN.** NaN poisons through `lerp`;
  infinity does not, and the shader already clamps an infinite corner radius
  (`min(radius, min(half.x, half.y))`), which makes it a plausible spelling for
  a pill. Refusing both is simpler; refusing only NaN is narrower.

### Done when

A signal that goes non-finite makes its widget stop moving, says so once, and
starts following again when the signal recovers — without the surface having
spent the interval rendering at the frame rate.

MalpenZibo:
## What changed

A declared value that is not a number arithmetic can carry on with is now
coerced **where it is resolved**, once, and reads as if it had not been
declared. Nothing downstream — the matrix, the animation, hit testing, the
damage rect — needs a guard, because none of them can be handed one any more.

That placement is the whole design, and it was chosen after prior art rather
than by preference. CSS coerces a NaN escaping a `calc()` at computed-value
time so no property ever sees one; Web Animations falls back to the identity
easing; Flutter asserts in debug; Qt warns. Nobody lets it reach the matrix
silently, and the *one that fixes all three symptoms at once* is CSS's, because
downstream never sees a bad number to be poisoned by.

The three symptoms #228 names, and why one door closes all of them:

- **The matrix.** `sin_cos(NaN)` makes the whole affine NaN, `is_identity`
  compares equal to nothing so it is applied, and `inverse`'s
  `det.abs() < 1e-10` is false for NaN so hit testing returns NaN coordinates.
- **The animation, permanently.** `lerp` from a NaN start is NaN at every `t`,
  so a later and perfectly finite target animates from nothing to nothing for
  the life of the process. Confirmed against `AnimationState` directly; the
  widget-level path masks it, which is why the recovery test here guards
  recovery rather than proving the poisoning.
- **The frame loop.** `animate_to`'s early-out is `new_target == self.target`,
  and NaN is equal to nothing including itself — so every pass began a segment
  and asked for another Animation job, at the compositor's rate, for ever.

`AllFinite` deliberately does **not** hang off `Animatable`: "can this be
interpolated" and "is this a number" are different questions, and the first
draft that conflated them could not spell the door for `Length` or `Pivot` —
which is most of the ways a bad number gets in. The animatable types answer
from `channels()`, which they already provide, so the check cannot go stale
when a field is added.

Closes #228.

## What proves it

Six tests in `src/widgets/container/characterization.rs`, each failing without
the change:

- `..._reads_as_if_it_were_not_declared` — the coerced value is the property's
  own default, not NaN and not "somewhere".
- `..._lets_the_surface_go_idle` — the battery symptom.
- `a_property_follows_its_signal_again_once_the_signal_recovers` — against a
  control that was never given a bad number, so the assertion is "these two
  agree" rather than a restatement of which way round the matrix stores a sine.
- `a_pivot_that_is_not_a_number_does_not_swallow_a_click` — the pivot reaches
  `center_at` with no state layer on its way, and what a bad one costs is not a
  wrong picture but a **dead widget**: every hit test answers no and the
  container silently stops taking clicks plainly inside it.
- `..._is_reported_once` — five paints, one message.
- `every_declared_property_is_resolved_to_a_finite_value` — the door itself.

That last one is the load-bearing test and it took three attempts to write
honestly. It asks each resolver for the value it hands out, because the two
versions before it asserted on painted output and **passed with the doors
removed** — a NaN width is clamped into a finite size by the layout beneath it,
and a NaN shadow never reaches a drawn rect. I verified the current one fails
when each of `shadow`, `shadow extent`, `width` and the state-layer filter is
taken away.

No golden or snapshot moved, and none should: the change alters behaviour only
when a declared value is non-finite, which no existing scenario declares.

## What the review found

`/simplify` ran first, four readings. Its altitude pass found the blocking
problem of this change: **the door had seven holes.** `shadow` and `padding`
caused exactly the never-settles bug the issue is named after; `max_transform_reach`
fed `Transform::compose` three raw values; `width`/`height` could not use the
door at all. All closed, and the predicate moved off `Animatable` so
non-animatable types can reach it. Its efficiency pass measured the guard as
**free in release** — the `SmallVec` disappears and the check auto-vectorizes;
hand-writing per type is 52 instructions against 47 and no SIMD. In debug it is
~10× a hand-written check, with no fix available (no inliner at `-O0`); worth
knowing before anyone profiles a debug build, not worth a change.

The `reviewer` subagent then ran over the commits. **One blocking finding, fixed:**
`shadow` has a *second* resolver, `max_shadow_extent`, still reading raw — so a
container with a bad shadow and a transform published a damage rect that did not
cover where it drew. Closed, and the enumeration test now asks it.

Its three "worth answering" findings are all acted on: the test now asserts
`width`, `height`, the shadow extent and a state-layer override; the commits
were re-split so each compiles under `-D warnings` alone (the first version had
commit 1 calling a symbol introduced in commit 2, which I confirmed by checking
it out); and the `Text`/`TextInput` residue is filed as #333 rather than left in
this body. Its notes are done too — `finite` is `pub(crate)` rather than public
API nobody can use, and two `track_caller` attributes that read no location are
gone.

## What was checked by hand

Nothing on a compositor, and none needed: every claim here is a number a test
reads back. Two things were measured rather than reasoned about — the animation
poisoning, probed directly against `AnimationState` because the widget path
masks it, and each door removal, to confirm the enumeration test actually fails.

## Left undone

**#333 — `Text` and `TextInput` resolve their font size with a plain read**, so
a non-finite one still reaches `animate_to` and poisons that animation for good.
It is the one residue that meets the bar: reachable with
`text("x").font_size((move || w.get() / n.get()).transition(200.0))`, a caller is
worse off today, and its acceptance criterion is one this change's tests cannot
state — `TextStyle` has no state-layer resolution, so there is no
`resolve_state_value` to hang a door on and the widgets need one built.

Three more unguarded reads were checked and **do not** earn an issue, which is
the point of checking: gradients and ripple colour are paint-only and reach the
shader rather than the matrix, layout, hit testing or the damage rect;
`Scroll`'s metrics are unguarded but `(max_width - gutter).max(0.0)` turns the
NaN into a zero before it reaches geometry. Nobody is worse off today in a way I
can name.

The substitute is the property's **own default** — identity for every transform
component — rather than the last finite value it held. That is CSS's semantics
and needs no per-property storage; holding the last value would mean a `Cell`
per property and was the alternative I would otherwise have recommended. It
means a widget rotated 45° snaps upright rather than freezing at 45°, which is
visible, defined, and cannot be mistaken for still following.

https://claude.ai/code/session_016cdRGJettCQPPACDkS1CtM

### DIFF
--- a/src/animation/animatable.rs
+++ b/src/animation/animatable.rs
@@ -54,7 +54,7 @@ pub trait Animatable: Copy + PartialEq + Send + Sync + 'static {
 
     /// The value as the vector of numbers the animation interpolates.
     ///
-    /// Only `carry_velocity` reads this, and only as a direction: a spring's
+    /// `carry_velocity` and the finiteness check read this, and only as a direction: a spring's
     /// momentum is a vector in this space, and what a new segment inherits is
     /// the part of it pointing along itself. The channels may be in different
     /// units — `Corners` carries four radii in pixels beside a unitless
--- a/src/finite.rs
+++ b/src/finite.rs
@@ -128,6 +128,9 @@ impl AllFinite for crate::pivot::Pivot {
             V::Percent(p) | V::Px(p) => p.is_finite(),
             _ => true,
         };
+        // Both arms, and both are watched: the enumeration test puts the bad
+        // number in the horizontal anchor and the hit-testing one puts it in
+        // the vertical, so neither can be deleted unnoticed.
         horizontal && vertical
     }
 }
--- a/src/finite.rs
+++ b/src/finite.rs
@@ -1,0 +1,133 @@
+//! Whether a declared value is a number arithmetic can carry on with.
+//!
+//! A property is a closure the application wrote, and a division by a measured
+//! zero is all it takes to produce a NaN. Once one is in it stays: `lerp` from
+//! a NaN start is NaN at every `t`, so a later and perfectly finite target
+//! animates from nothing to nothing for the life of the process — and
+//! `animate_to`'s `new_target == self.target` early-out never fires, because
+//! NaN is equal to nothing including itself, so the surface asks for another
+//! frame at the compositor's rate for ever.
+//!
+//! The answer is CSS's: a value that is not a number is coerced *where it is
+//! resolved*, once, so no consumer downstream ever has to know it happened. The
+//! alternative — a guard at each consumer — is what was tried before, and each
+//! guard changed what the next one saw.
+//!
+//! This lives apart from [`Animatable`](crate::animation::Animatable) on
+//! purpose. "Can this be interpolated" and "is this a number" are different
+//! questions, and hanging the second on the first put the door out of reach of
+//! every declared value that is not animatable — a [`Length`], a [`Pivot`], a
+//! gradient — which is most of the ways a bad number gets in.
+//!
+//! [`Length`]: crate::layout::Length
+//! [`Pivot`]: crate::pivot::Pivot
+
+use crate::reactive::Signal;
+
+/// A value that can say whether every number inside it is finite.
+///
+/// Implemented for every type a property can be declared with. The name is
+/// `all_finite` rather than `is_finite` because `f32` already has an inherent
+/// method by that name: a trait method taking `&self` matches a `&f32` receiver
+/// *exactly* while the inherent one needs a deref, so `c.is_finite()` inside a
+/// blanket body resolves to the trait and calls itself until the stack ends.
+pub(crate) trait AllFinite {
+    /// Whether this value is one arithmetic can carry on with.
+    fn all_finite(&self) -> bool;
+}
+
+impl AllFinite for f32 {
+    fn all_finite(&self) -> bool {
+        f32::is_finite(*self)
+    }
+}
+
+/// Reading a declared property, with a value nothing can compute treated as no
+/// declaration at all.
+pub(crate) trait FiniteOr<T> {
+    /// The value, or `default` where it is not finite.
+    ///
+    /// `property` is only for the diagnostic, and is the name a caller would
+    /// recognise — the setter they wrote, not the accessor that resolves it.
+    fn get_finite_or(&self, default: T, id: crate::tree::WidgetId, property: &'static str) -> T;
+
+    /// The same, for the paths that read a snapshot rather than subscribing —
+    /// the reach calculation, which runs beside a paint that is reading the
+    /// same signal and has already reported it.
+    fn get_finite_or_untracked(&self, default: T) -> T;
+}
+
+impl<T: AllFinite + Clone + 'static> FiniteOr<T> for Option<Signal<T>> {
+    fn get_finite_or(&self, default: T, id: crate::tree::WidgetId, property: &'static str) -> T {
+        let value = crate::reactive::OptionSignalExt::get_or(self, default.clone());
+        if value.all_finite() {
+            value
+        } else {
+            crate::reactive::diagnostics::non_finite_value(id, property);
+            default
+        }
+    }
+
+    fn get_finite_or_untracked(&self, default: T) -> T {
+        let value = crate::reactive::OptionSignalExt::get_or_untracked(self, default.clone());
+        if value.all_finite() { value } else { default }
+    }
+}
+
+/// Every type a `Container` property is declared with, answered from the
+/// numbers it is made of.
+///
+/// The animatable ones already have to say which numbers those are in order to
+/// be interpolated at all, so they answer from
+/// [`channels`](crate::animation::Animatable::channels) and cannot go stale
+/// when a field is added — the last attempt at this wrote the check out per
+/// type, covered three of seven, and nobody noticed the other four.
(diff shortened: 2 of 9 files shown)


MalpenZibo:
Mutation job on the first push: **51 mutants, 42 caught, 3 missed, 6 unviable.** The one that was this change's own is fixed in `9752f0b` — deleting the vertical arm of `Pivot::all_finite` killed no test, because both pivot scenarios put the bad number in the horizontal anchor. They now take one each.

The two that survive are pre-existing gaps on lines this diff touched, recorded rather than filed:

- `anim_bridge.rs:202` — inverting the padding drift comparison (`==` → `!=`) kills nothing, so nothing watches padding's drift check. The line changed here only because `padding` gained a single resolver; the gap predates it.
- `diagnostics.rs:165` — `imp::reset` can be replaced with `()`. Test-support plumbing, and this change only added a line to it.

Neither has a caller who is worse off, so by AGENTS.md's bar they are a sentence rather than an issue.

MalpenZibo:
Merging on the blocking checks, all green. The mutation job runs 17 minutes and is reporting-only, so rather than spend a third round on it I deleted each surviving arm by hand and watched the tests fall over — which is what the job was going to do.

Final state of its three survivors from the first run:

- `finite.rs` pivot arms — **fixed and verified locally.** The first attempt only moved the gap: both scenarios had a bad number in the *other* anchor, and a case with both anchors bad cannot kill either arm, because whichever survives still detects it. Each test now takes one anchor.
- `anim_bridge.rs:202` and `diagnostics.rs:165` — pre-existing gaps on lines this diff touched, left as a sentence rather than an issue: neither has a caller who is worse off.

PR #343, opened 2026-09-09
A conversion is declared once, and every form comes from that declaration

### LINKED ISSUE #226: A signal converts only for the pairs somebody remembered to list
`IntoSignal` has three forms — a value, a closure, a signal — and a property is
only properly reactive if the same expression compiles in all three. The value
and closure forms share one set of conversions, the `From` and `IntoVal` impls
that sit beside each type. The signal form has its own, and until #225 it had
none at all, so this was the state of things:

```rust
container().width(100.0)              // ok, From<f32> for Length
container().width(move || w.get())    // ok, IntoVal<Length> for f32
container().width(w)                  // Signal<f32>: did not compile
```

#225 added the missing impls for every pair that had an `IntoVal` — `Length`,
`Padding`, `Corners`, `BackdropBlur`, `Scale`, `Translate`, the numeric
widenings into `f32`, and the optional case. What it could not add is a rule.

### Why it is a list and not one impl

The obvious blanket does not compile:

```rust
impl<S, T> IntoSignal<T, ConvertedSignalMarker> for Signal<S>
where S: IntoVal<T> { .. }
```

`IntoVal` is reflexive — `impl<T> IntoVal<T> for T` — so this also covers
`Signal<Length> -> Length`, which the passthrough impl already covers with a
different marker. Both apply, `M` cannot be inferred, and existing call sites
stop compiling. Measured: adding it breaks `container().rotate(f)` on a
`Signal<f32>` with `E0283 type annotations needed`.

Excluding the reflexive case needs either negative bounds or specialisation,
neither of which is stable. Naming the source type per impl excludes it by
construction, which is what `converting_signals!` does — at the cost of being a
list somebody has to remember to extend.

The one case that *does* generalise is `Option<T>` from `T`, because reaching
`Option<X>` from a `Signal<Option<X>>` is the passthrough and the converting
impl is only ever selected for a `Signal<X>`. That one is written as a blanket.

### What is left

**The drift.** A new type with a `From` gets the value and closure forms from
one `IntoVal` impl and the signal form only if someone remembers
`converting_signals!`. Nothing fails at compile time; the gap shows up as a
user's compiler error months later. That is exactly how `.width(signal)` stayed
missing.

**Pairs not covered**, because it was not clear they are wanted rather than
because they are hard:

- `Signal<String>` / `Signal<&'static str>` into `ImageSource`, which has
  `From<&str>`, `From<String>` and `From<PathBuf>`
- `Signal<[f32; 2]>` and friends into `Padding`, which takes the arrays as
  values but not yet as signals
- `Signal<Color>` into `Option<Color>`-shaped properties beyond the gradient

**A test that would catch it.** `tests/signal_conversions.rs` is a list of
spellings, so it only covers what somebody thought to write down. What would
actually hold the line is a check that every `IntoVal<T> for S` has a matching
`Signal<S> -> T` — a macro that emits both from one declaration, or a build-time
assertion over the pairs. Emitting both from one place looks like the tractable
version: `IntoVal` and the converting signal impls would stop being two lists
that can disagree.

### Not urgent

Everything has a working spelling today — `move || sig.get()` — and it is one
closure. This is about the API being predictable rather than about anything
being unreachable.

MalpenZibo:
## What changed

A property's conversions were two lists kept in step by hand — an `IntoVal<To> for From` beside each type, and a `converting_signals!(From => To)` invocation repeating the same pairs — and nothing enforced the mirror, so a pair added to one and not the other compiled as a closure and refused as a signal at somebody else's call site months later. There is now one `converts!` macro and it emits every form from a single declaration. The `=>` shape calls `<To>::from`, which will not compile unless the `From` beside the type exists, so the value form is proved rather than mirrored; the `as` shape is a widening std has no `From` for, and there the macro emits the value form too. The same 52 pairs convert as before. Closes #226.

## What proves it

`one_declaration_covers_every_form` in `src/reactive/into_signal.rs`, two tests over two throwaway pairs — `f32 => Millimetres` for the `=>` shape and `i64 as f32` for the `as` shape — each exercising all five spellings a property accepts. Both fail to compile before the macro exists, for the stated reason: the value form arrives through `From` and the other four do not exist.

The review deleted each of the five emitted impls in turn and confirmed every deletion breaks the build with these tests among the failing sites, the `Memo` and `Signal` cases failing **only** there. I ran the same experiment on the `LossyMarker` impl the `as` arm now emits, and the failure names `i64`.

`tests/signal_conversions.rs` is the caller-facing list and does not shrink. Full suite 980 passing, clippy clean, goldens 10 passing on lavapipe, `cargo doc`, `mdbook build` and `mdbook test` clean. No golden moved and none should: this touches no renderer code.

## What the review found

Zero blocking. Three worth answering, all acted on:

- A commit message called the `u16` case drift. It is the opposite — std has `From<u16> for f32`, so a second value impl would make `4u16` ambiguous between the two markers, and `u16` belongs in the `=>` shape for that reason. The message now says so, since a wrong sentence there outlives the change.
- The same message claimed a byte-level size win with no reproducible method. Removed rather than dressed up.
- The reviewer's own rule described only the `=>` shape, so a reviewer applying it would ask an author for a `From` that an `as` entry must not have. Both shapes are spelled out now.

Its notes are also applied: the macro's comment says the two shapes cannot be mixed in one invocation and that `@pair` is internal, and the stale "three forms" wording in `tests/signal_conversions.rs` is now five.

Before that, four cleanup readings found the macro was a token-muncher where all ten other macros in the crate are flat repetitions, and that the three `LossyMarker` impls were themselves a hand-kept mirror of the `as` list — the very defect this closes, sitting twenty lines above the macro, and already out of step with it. Both are fixed in the commits rather than on top of them.

## What was checked by hand

Nothing. The harness covers all of it. Sixty examples exercise the value form with bare literals and compile as part of clippy's `--all-targets`.

## Left undone

The pairs #226 lists as uncovered stay uncovered, deliberately: `ImageSource` from a `String` or a `&str`, and `Color` into optional properties beyond the gradient. `ImageSource` is worse than the issue says — it has three `From` impls and no `IntoVal` at all, so its *closure* form is missing too, not only its signal form. That is a missing pair rather than a missing mechanism, it is separately provable now that one declaration exists, and it stays on #226's successor rather than being smuggled in here.

`converts!` is `pub(crate)`, so a widget crate outside guido still cannot declare a conversion pair, even though `IntoVal` is public and in the prelude for exactly that audience. Unchanged by this diff and out of scope for #226 — exporting a macro is a public API decision. No issue opened: I cannot name somebody worse off today rather than somebody who might be.

https://claude.ai/code/session_01Q5KL58bdTdT533XJa1ZdDn

### DIFF
--- a/.claude/agents/reviewer.md
+++ b/.claude/agents/reviewer.md
@@ -1,6 +1,6 @@
 ---
 name: reviewer
-description: Reviews a change against Guido's own rules — harness coverage, the reactivity rule, the three spellings, atomic commits, documentation. Use before opening a pull request, or on an existing PR diff. Reports findings; does not fix them.
+description: Reviews a change against Guido's own rules — harness coverage, the reactivity rule, the five spellings, atomic commits, documentation. Use before opening a pull request, or on an existing PR diff. Reports findings; does not fix them.
 tools: Read, Grep, Glob, Bash
 ---
 
@@ -42,9 +42,14 @@ them.
 
 - **The reactivity rule**: anything that survives to paint takes
   `impl IntoSignal<T, M>`; structural declarations do not.
-- **The three spellings**: a new conversion needs `From`, `IntoVal` *and*
-  `converting_signals!`, plus a line in `tests/signal_conversions.rs`. A missing
-  signal form compiles fine and refuses at a call site months later.
+- **The five spellings**: a new conversion is one `converts!` entry, which
+  emits the closure and signal forms together, plus a line in
+  `tests/signal_conversions.rs`. An `S => T` entry needs the `From` beside the
+  type, and will not compile without it; an `S as T` entry is a widening std
+  has no `From` for, where the macro emits the value form too and a `From`
+  would make the marker ambiguous. An `IntoVal` impl written by hand
+  is the drift #226 closed: it gives the closure form and not the signal one,
+  which compiles fine and refuses at a call site months later.
 - Position and bounds live in the `Tree`, never on the widget.
 - Performance claims come with a measurement, before and after.
 - Commits are atomic and their subjects are sentences saying what is now true.
--- a/.claude/skills/widgets/SKILL.md
+++ b/.claude/skills/widgets/SKILL.md
@@ -57,36 +57,50 @@ than a value quietly ignored.
 A new property lands on one side of that line. If it changes what a pixel looks
 like, it is reactive.
 
-## A reactive property has three spellings, and they must agree
+## A reactive property has five spellings, and they must agree
 
-`IntoSignal` accepts a value, a closure, or a signal. A property is only
-properly reactive when the *same expression* compiles in all three positions:
+`IntoSignal` accepts a value, a closure, a signal, a writable signal or a memo.
+A property is only properly reactive when the *same expression* compiles in
+every position:
 
 ```rust
 container().width(100.0)             // value
 container().width(move || w.get())   // closure
-container().width(w)                 // signal
+container().width(w)                 // signal, writable signal or memo
 ```
 
-Three different sets of impls serve them, and they drift apart silently —
-nothing fails to build when one is missing; a call site refuses months later.
+**A conversion is declared once, and `converts!` emits every form:**
 
-**Adding a conversion to a property type means adding all three:**
+```rust
+crate::reactive::converts!(
+    f32 => Length,
+    f64 => Length,
+);
+```
 
-| spelling | impl | where |
-| --- | --- | --- |
-| value | `From<S> for T` | beside `T` |
-| closure | `IntoVal<T> for S` | beside `T` |
-| signal | `converting_signals!(S => T)` | beside `T` |
+That writes `IntoVal<Length> for f32` — the closure form — and the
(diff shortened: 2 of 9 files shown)


MalpenZibo:
CI is green on all eight checks, but the mutation job is worth reading rather than counting: it reports `No mutants to filter`.

That is not a clean sweep, it is an empty one. Everything this change adds lives inside a `macro_rules!` body, and `cargo mutants` does not generate mutants for macro definitions — the impls it would want to mutate do not exist as source lines until expansion. The rest of the diff is deletions, which have nothing to mutate either. So the job asked its question of nothing.

What stands in its place is the deletion experiment recorded under **What proves it**: each of the five impls the macro emits was removed in turn and the build was confirmed to break with the new tests among the failing sites, the `Memo` and `Signal` cases failing only there. That is the same question the mutation job asks — if this were wrong, would anything notice — answered by hand because the tool cannot reach here.

Worth knowing generally: any future change whose substance is inside a macro gets the same empty pass from that job.

Issue #344, opened 2026-09-09
Nothing tests whether a blur region ever reaches the platform, only what it would contain


MalpenZibo:
## Observed

`regions_from_commands` is well covered — its shape, its clipping, its transforms, and since #268 its radius. What nothing covers is the hop *after* it: whether the region is handed to the platform at all.

`src/lib.rs:1792-1795`:

```rust
let blur_rects = blur::regions_from_commands(&surface.flattened_commands);
```

reached only when `flatten_root_into`'s `compositor_blur` return, or `handle.has_published_blur()`, says so. That disjunction is the switch deciding whether `sync_blur_region` is called on a given frame, and the `has_published_blur()` half of it is what withdraws a region that has just gone away.

The container harness cannot see it. `src/widgets/container/characterization.rs:113`:

```rust
let _ = crate::renderer::flatten_root_into(&node, &mut commands, &mut layers);
let blur = crate::blur::regions_from_commands(&commands);
```

The return is discarded and `regions_from_commands` is called unconditionally, so every test in that file asks "what would the region be" and none asks "was it published". A change that made `compositor_blur` always `false` would leave the whole blur suite green and ship a build where no compositor blur is ever requested.

## Who is worse off

Anyone on a compositor with `ext-background-effect-v1`. The failure is silent in exactly the way #268 was: the region is computed correctly, nobody sends it, and what appears on screen is the compositor's own fallback rule.

## Expected

A test that fails when the publish decision is wrong, not merely when the region's contents are.

## Acceptance

`tests/headless_app.rs` — the recorder already answers for every surface the loop holds, and the loop is what calls `sync_blur_region`. A surface whose container asks for a compositor blur records one region set; the same surface after the blur is withdrawn records the withdrawal on that frame and nothing afterwards. Both fail if `compositor_blur` is forced to `false`, and the second fails if `has_published_blur()` is dropped from the disjunction.

That is the acceptance #268's tests could not state: they stop at the command list by construction, because the harness that runs them never drives the loop.

## Where it lives

`src/lib.rs:1745-1795` — the two call sites and the guard.
`src/renderer/flatten.rs:355` — where `compositor_blur` is set.
`src/platform/backdrop.rs:94` — `sync_blur_region`, and `:113` where it diffs against the last published region.
`src/widgets/container/characterization.rs:113` — the harness that discards the flag.

## Prior art

None needed. `tests/headless_app.rs` already exists for exactly this layer, and #264 built the recorder that makes it reachable.

---

[...]

PR #345, opened 2026-09-09
A blur is asked for by what it says, not by a radius the other half never reads

### LINKED ISSUE #268: A blur only the compositor performs is dropped for having no radius of its own

MalpenZibo:
## What changed

A container that asked only the compositor to blur its backdrop emitted no draw command, so nothing downstream could recover: no command meant no `compositor_blur` after flattening, so `sync_blur_region` was never called, so the region pass that correctly tessellates the rounded shape was never reached. The guard was `blur.radius > 0.0`, and that radius belongs to the blur guido runs in its own shader — the documentation beside `BackdropSources::COMPOSITOR` says in as many words that it does not apply there, because the protocol carries no radius. The rule now lives beside that documentation: where the surface is a source the radius decides and zero means off, otherwise the sources decide alone. Closes #268.

## What proves it

`a_compositor_blur_needs_no_radius_of_its_own` in `tests/render_snapshots.rs`, which is the acceptance criterion verbatim — two containers, a `BackdropBlur` command for the compositor-only one and none for the both-sources one. It is a new scenario with a new reference, so no existing snapshot was rewritten and no `golden-update` label is needed.

Beside it, `a_compositor_only_blur_is_published_without_a_radius` and `an_emptied_source_set_asks_for_nothing_at_any_radius` in the container characterization, which drive a real frame. All three fail with the guard reverted; the review confirmed that independently by reverting it.

`a_compositor_region_does_not_need_a_radius` in `src/blur.rs` is the unit test the issue asked for, and it passes with or without the fix — `regions_from_commands` was always correct, and this is the first test it has ever had. It is regression coverage for the chain, not proof of the change, and should not be counted as such.

No golden moved, which is the right outcome rather than a missing scenario: a compositor-only command is routed to the Shapes layer, produces no instance, and `command_to_backdrop_region` returns `None` without `SURFACE`, so it draws no pixels guido controls.

## What the review found

Zero blocking. Both findings worth answering are acted on:

- The divergence between the issue's **Expected** and **Acceptance** sections was recorded only in this change, not where somebody would look for it. It is now a comment on #268 saying which reading was implemented and why the other would have broken two existing tests.
- `docs/STYLING.md` and the styling chapter taught the zero rule unqualified, and that is where a reader learns it. Both now name the exception.

Before that, four cleanup readings found the predicate I first wrote was an instance of the rule rather than the rule: `BackdropSources::empty()` at a non-zero radius returned true from a method called `asks_for_anything`, which contradicted the off switch the same change was documenting. Fixed in the commit, and `an_emptied_source_set_asks_for_nothing_at_any_radius` is what holds it — it fails against the first spelling.

Two notes not acted on. The efficiency reading found the new command adds one rect to the Shapes layer bounds and makes frames run the region scan that previously skipped it; both are what a compositor blur at radius 16 already paid, so neither is new machinery. And `draw_text_backdrop_blur` keeps its equivalent guard inside the paint context while `draw_backdrop_blur` leans on its single caller — two altitudes for one kind of decision, with nobody worse off while the container is the only caller.

## What was checked by hand

**Nothing, and this is the one claim that is unproven at the layer the bug was seen.** #268 opened with a `WAYLAND_DEBUG=1` table from niri 26.04 showing zero `get_background_effect` requests at radius `0.0`. The matching after-row has not been run, because doing so opens a surface on the maintainer's screen. Everything up to and including the region contents is covered automatically; the last hop, the flag reaching `sync_blur_region`, is watched by nothing — which is #344.

I can run `WAYLAND_DEBUG=1 cargo run --example blur_example` to a log file on request and report the row.

## Left undone

`BackdropBlur::new(24.0).sources(BackdropSources::empty())` used to emit an inert command that neither consumer read, and now emits nothing. No pixels and no region either way, it has its own test, and it is what makes the documented off switch true — named here so it is not discovered later as an unexplained change.

**#344** — nothing tests whether a blur region ever reaches the platform, only what it would contain. The container harness discards `flatten_root_into`'s `compositor_blur` return and calls the region pass unconditionally, so forcing that flag to `false` would leave the entire blur suite green. That needs `tests/headless_app.rs` and the real loop, which is an acceptance criterion this change's tests cannot state by construction.

https://claude.ai/code/session_01Q5KL58bdTdT533XJa1ZdDn

### DIFF
--- a/book/src/advanced/wayland.md
+++ b/book/src/advanced/wayland.md
@@ -664,6 +664,26 @@ small rectangles because `wl_region` has no notion of curves, and the
 protocol carries no radius — the compositor picks its own, so `radius`
 does not apply there. Check availability with `compositor_effects()`.
 
+Which is why a blur restricted to the compositor asks for its region on
+its sources alone, and a radius of zero there is a request rather than a
+refusal:
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+// Blur nothing of our own; let the compositor blur the desktop behind us.
+container().backdrop_blur(
+    BackdropBlur::new(0.0).sources(BackdropSources::COMPOSITOR),
+)
+# ;
+# }
+```
+
+Everywhere the surface is a source, a radius of zero still means no blur
+— that is how a blur is switched off by the same signal that switches it
+on. To switch off a compositor-only one, empty its sources instead.
+
 The **surface** side is guido's own: the frame is drawn into an
 offscreen target, each region is downsampled, blurred with a separable
 gaussian and composited back masked to the container's rounded shape.
--- a/book/src/building-ui/styling.md
+++ b/book/src/building-ui/styling.md
@@ -220,7 +220,10 @@ container()
 
 A blur radius of `0.0` is "no blur", on a container and on a text alike, which
 is what lets one signal switch the effect on and off rather than forcing the
-caller to rebuild the widget in a Rust branch.
+caller to rebuild the widget in a Rust branch. That holds wherever the surface
+is one of the sources, which is everything by default. A container that
+restricts itself to the compositor has no radius of its own — the protocol
+carries none — so it asks on its sources alone, and empties them to switch off.
 
 **What is not reactive** is structural: `.layout(..)`, the axis a `.scroll(..)`
 is built with, `.control()`, and the motion a
--- a/docs/STYLING.md
+++ b/docs/STYLING.md
@@ -239,7 +239,10 @@ container()
 
 A blur radius of `0.0` is "no blur", on a container and on a text alike, which
 is what lets one signal switch the effect on and off rather than forcing the
-caller to rebuild the widget in a Rust branch.
+caller to rebuild the widget in a Rust branch. That holds wherever the surface
+is one of the sources, which is everything by default. A container that
+restricts itself to the compositor has no radius of its own — the protocol
+carries none — so it asks on its sources alone, and empties them to switch off.
 
 **What is not reactive** is structural: `.layout(..)`, the axis a `.scroll(..)`
 is built with, `.control()`, and the motion a value is declared with —
--- a/src/backdrop.rs
+++ b/src/backdrop.rs
@@ -74,6 +74,19 @@ impl BackdropBlur {
         self.sources = sources;
         self
     }
+
+    /// Whether this request asks for anything at all.
+    ///
+    /// Where the surface is a source the radius decides, and zero means off.
+    /// Otherwise the sources decide alone, for the reason
+    /// [`BackdropSources::COMPOSITOR`] gives: that half carries no radius.
+    pub(crate) fn asks_for_anything(&self) -> bool {
+        if self.sources.contains(BackdropSources::SURFACE) {
+            self.radius > 0.0
+        } else {
+            !self.sources.is_empty()
+        }
+    }
 }
 
 impl From<f32> for BackdropBlur {
--- a/src/blur.rs
+++ b/src/blur.rs
@@ -218,6 +218,37 @@ mod tests {
         EllipticalRadii::circular(CornerRadii::uniform(radius))
     }
 
+    /// The radius belongs to the blur guido runs in its own shader. The
+    /// compositor picks its own, so a region is published for a compositor
+    /// source whatever the radius says — including zero, which is the spelling
+    /// for "blur nothing yourself, let the compositor blur what is behind me".
+    #[test]
+    fn a_compositor_region_does_not_need_a_radius() {
+        let command = FlattenedCommand {
+            command: std::rc::Rc::new(DrawCommand::BackdropBlur {
+                rect: Rect::new(0.0, 0.0, 100.0, 50.0),
+                sources: BackdropSources::COMPOSITOR,
+                radius: 0.0,
+                corner_radii: CornerRadii::uniform(0.0),
+                curvature: Default::default(),
+            }),
+            world_transform: Default::default(),
+            world_transform_origin: None,
+            layer: Default::default(),
+            clip: None,
+            clip_is_local: false,
+        };
+        assert_eq!(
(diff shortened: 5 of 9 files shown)


MalpenZibo:
## The wire row, run

The claim the body left unproven is now proven. `WAYLAND_DEBUG=1` on `examples/blur_example`, which asks for `BackdropBlur::new(0.0).sources(BackdropSources::COMPOSITOR)` — the exact spelling that said nothing for the life of the process before this change.

| `BackdropBlur::new(0.0).sources(COMPOSITOR)` | before (from #268) | after |
| --- | --- | --- |
| `get_background_effect` | 0 | 1 |
| `wl_region.add` | 0 | 49 |
| `set_blur_region` | 0 | 1 |

```
-> ext_background_effect_manager_v1#17.get_background_effect(new id ext_background_effect_surface_v1#61, wl_surface#18)
-> wl_compositor#3.create_region(new id wl_region#62)
-> wl_region#62.add(58, 57, 304, 1)
-> wl_region#62.add(51, 58, 318, 1)
-> wl_region#62.add(48, 59, 324, 1)
   ... 46 more
-> ext_background_effect_surface_v1#61.set_blur_region(wl_region#62)
```

The 49 rects are the rounded rect tessellated, and they read as one: each is a one-pixel-tall row, and the rows widen from 304 to 318 to 324 to 328 as they descend, which is the corner curve opening out. That is `rounded_rect_to_blur_rects` doing what it always did correctly and never got the chance to do, because the command it reads was never emitted.

Before this change the client bound the manager, received `capabilities(1)`, and then said nothing at all. The whole chain now runs end to end: the guard emits the command, flatten sets `compositor_blur`, `sync_blur_region` is reached, and the region goes out on the wire with the container's own shape.

Still true that nothing automated watches this row — that is the table in `AGENTS.md`, and #344 covers the hop below it.

Issue #347, opened 2026-09-10
Padding and border width never report drift, so a write that lands after their subscription is lost


MalpenZibo:
## Observed

`resync_animation_targets` is the pull half of the invariant the `anim_bridge` module header describes: every paint, re-read each animated target and ask for an Animation job if any has drifted from what its animation is aiming at. It is what converges a write that landed before the subscription existed.

Every property it checks is gated on the animation having been initialised — `src/widgets/container/anim_bridge.rs:216-232`:

```rust
let moved = |initial: bool, same: bool| !initial && !same;
```

`initialized` is set in exactly one place, `AnimationState::set_immediate`. And **nothing has ever called `set_immediate` for padding or border width.** They are seeded at builder time by `declare`, from `signal.get_untracked()`, and `seed_animations` reads their targets only for the subscription — the module comment says so in as many words: "those two are read there only for their subscription, so their seed survives into the first advance".

So `is_initial()` stays true for both, `moved(..)` is false for both, and neither has ever reported drift.

## Who is worse off

An application whose padding or border width is written between the container being built and its first paint — the same window the whole pull half exists for. The value is read once at construction and the write is lost until something else happens to wake that widget. It is the defect `resync_animation_targets` was written to close, still open for two of the eleven properties.

## Expected

Padding and border width converge like every other animated property, or the gate stops being keyed on a flag they can never set.

## Acceptance

In `src/widgets/container/characterization.rs`, beside the tests that already exercise the resync path: a container with an animated padding whose signal is written after construction and before the first paint arrives at the written value. It fails today. The same for border width.

Whichever way it is fixed, the pull half has to be watched for these two — the current state is that deleting their branch from `resync_animation_targets` changes nothing.

## Where it lives

`src/widgets/container/anim_bridge.rs:216-232` — the gate.
`src/widgets/container/anim_bridge.rs:112-160` — `seed_animations`, where the other nine are placed and these two are only subscribed.
`src/widgets/container/animations.rs` — `set_immediate`, the only writer of `initialized`.

## Prior art

None needed. This is an invariant the module already states failing to hold for two of its members.

---

Found by the review of the pull request closing #303. Not caused by it: #303 added the first call that touches these two at seed time, which is what made the gap visible, and confirmed present before that change.

MalpenZibo:
## Correction, and what actually shipped

[...]

Issue #278, opened 2026-08-28
The mutation job spends nine tenths of its time linking, and a third of its mutants on code it never compiled


MalpenZibo:
## What was observed

The `Mutation testing` job on #277 took **30 minutes** and reported
`219 mutants tested: 185 missed, 7 caught, 27 unviable`. Run locally it is worse:
two attempts on a laptop, the first killed by `Disk quota exceeded` after 7
mutants, the second abandoned for making the machine unusable.

Two separate problems live in that one number. The job is slow, and the number
it reports is not information.

### It is 89% build, and the mutants are not the reason

Parsed from the job log of run 33162... (#277):

| | |
| --- | --- |
| build time, summed over mutants | **5085s** (85 min across 4 workers) |
| test time, summed over mutants | ~600s |
| build per mutant | min 1s, **median 36s**, max 338s |

Halving the mutant count would still leave a fifteen-minute job. The cost is
that **every mutant recompiles 48.6k lines of `guido` and relinks 22 test
binaries**, with the default `ld`: there is no `.cargo/config.toml` in the
repository, so nothing has ever pointed the build at `lld` or `mold`.

The comment on the job already records the last round of this — 26 GB and 84s
down to 1.4 GB and 47s, by dropping the examples and the debug info. That
attacked disk. This is the link.

### 29% of the mutants were in a file the job does not compile

63 of the 219 were in `src/testing.rs`. The job runs

```
cargo mutants --in-diff /tmp/pull-request.diff -C --lib -C --tests -j 4 --no-shuffle
```

with no `--all-features`, so the `testing` feature is off and `src/testing.rs`
is not in the build at all. Their build times say so plainly: **58 mutants
report `0s build`**. A mutation to code that is never compiled cannot be caught
by anything, and every one of them is reported as `MISSED`.

### The rest cannot die for want of an adapter

The job installs `libwayland-dev` and `libxkbcommon-dev` and no Vulkan ICD, so
every test that needs one skips. `tests/golden_images.rs`, the GPU unit tests and
now `tests/headless_app.rs` are all invisible to it. The `golden` job installs
`mesa-vulkan-drivers` and finds the lavapipe ICD in eleven lines; this job does
not. This was recorded as an asterisk on #274's number and is now the dominant
term: of the 106 `src/lib.rs` mutants reported missed, an unknown fraction is
killed by the headless tests on a machine that can run them.

**Unknown, and the point of the acceptance criterion below: nobody has ever run
[...]

PR #348, opened 2026-09-10
A value can say where it starts, not only how it moves

### LINKED ISSUE #303: A widget cannot appear mid-animation any more, and the three methods that let it were transform-only
## Observed

`animate_translate_from`, `animate_rotate_from` and `animate_scale_from` are
gone as of #215. They were the only way to say "on first layout, start
somewhere else and animate in", and nothing replaces them.

They were removed rather than ported because they are a different idea wearing
the `animate_` prefix. The other setters said *ease to the new value*; these
said *play a one-shot at birth, from a start value*. CSS keeps the two apart
too — `transition` for the first, `@starting-style` for the second. Porting
them into `Animated<T>` would have made `Motion` a three-way enum where two of
the three are the same idea and the third is not.

They were also transform-only. A card could slide or scale in; it could not
fade in, because there was no `animate_background_from`. That is the same
complaint #215 opens with, and re-spelling it would have carried it forward.

No caller outside the library used them: `examples/` did not, and the one
downstream consumer (`allock`) does not.

## Expected

One mechanism for entering, available for every animatable property, spelled
the way #215 spells the other two — riding with the value rather than declared
beside it. Something in the shape of

```rust
container().scale(Scale::NONE.entering_from(0.9, SpringConfig::SNAPPY))
```

with the same rule the rest of the vocabulary has: it is attached to a value
being set, so there is no way to spell one for a property that is not there,
and a state layer cannot declare one.

## Acceptance

- A container declared with an enter on any animatable property — a background
  and a scale at least, one paint-path and one transform — is mid-animation on
  the frame after its first layout, and at its declared value some frames
  later. Two containers differing only in whether the value carries an enter:
  one starts at the declared value, the other does not.
- Nothing plays a second time. An enter is consumed at the first layout, so a
  relayout, a resize or a state change does not replay it.
- `AnimationState` regains whatever it needs to hold a pending start value;
  #215 removed `enter_from`, `with_enter_from`, `take_enter_from`, `begin_from`
  and `seed_or_enter` when the last caller went, so this is a restoration with
  a wider reach rather than a new mechanism.
- The book documents it. `book/src/transforms/animated.md`'s API reference lost
  its enter section in #215 and gets one back that is not transform-shaped.

## Non-goals

- Exit transitions. `Transition::on_complete` is what "animate out, then
  destroy" is built on today and this does not change it.
- Making `Motion` public.

## Prior art

- **CSS `@starting-style`** — the styles an element has *before* its first
  style resolution, deliberately separate from `transition`. This is the shape
  the removal argued for.
- **Compose** `AnimatedVisibility` / `EnterTransition` — enter is a distinct
  vocabulary from `animate*AsState`, not a variant of it.

Opened from #215, which removed the old spelling. See the pull request
implementing it for what went and why.

### LINKED ISSUE #349: A negative font size hangs the process instead of drawing nothing
## Observed

One line of ordinary application code hangs guido, with no panic and no frame:

```rust
text("hello world this is a line").font_size(-1.0)
```

Reproduced in a throwaway integration test against `main` at `9ed91a6`, with the build completed first so the wait is the run and not the compile: laying that widget out at a 100x100 constraint does not return within 30 seconds. The same test with a positive size returns immediately.

It is not a slow path — nothing is drawn and no progress is made. I have not chased it past the layout call, so which loop spins is still open; the text measurer shaping to fit a `max_width` is the obvious suspect, since a non-positive advance never consumes the line.

## How it was found

The mutation job on #348. `cargo mutants` replaced `TextAnims::retarget` with `(None, -1.0)` and the mutant timed out at 30s rather than being caught — the only non-caught mutant of 51, and the reason that job is red. So the tool found it by accident: a negative size is exactly what that mutant fed to the measurer.

## Who is worse off

Anyone computing a font size rather than writing a constant. A size derived from a width divided by a count, from a configuration file, from a theme scale — any of them can go negative for the same reasons #333 says they can go non-finite, and the failure is worse: #333 poisons an animation for the life of the process, this one stops the process.

A bar that hangs takes the user's panel with it, and there is nothing on screen to say why.

## Expected

A non-positive font size draws nothing, or clamps, or panics with a sentence naming the value. Any of those is better than not returning. Whatever is chosen should match what #333 settles for a non-finite one — they are the same class and should not have two answers.

## Acceptance

An integration test that lays out a `text` with `font_size(-1.0)` and returns. It hangs today, so it needs a timeout guard or it will hang CI rather than fail it: assert on a value the layout produced, having run it on a thread with a join deadline, or clamp at the door so the layout is reached with a size that terminates.

Beside it, `0.0` and a very small positive size, which are the neighbours of the same boundary and are not covered either.

## Where it lives

`src/renderer/text_measurer.rs` — the shaping loop that most likely does not terminate.
`src/widgets/text.rs:204` and `src/widgets/text_input.rs:663` — where the size is resolved with `get_or(14.0)` and passed on unchecked, the same two sites #333 names.

## Prior art

Not needed for the shape — a size that cannot be drawn should not be handed to a loop that assumes it can. Worth checking what the answer is for #333 first, so the two agree.

---

Found by the mutation job on #348 and reproduced by hand. Not caused by that change: confirmed against `main` before it.

MalpenZibo:
## What changed

A widget can appear mid-animation again. `Animated::entering_from(value)` says where a property begins on the one frame its widget appears, composing with the `transition` that says how it travels — the split CSS makes between `@starting-style` and `transition`. It hangs off the value rather than off three transform setters, so unlike the `animate_*_from` methods #304 removed, a card can fade in. Closes #303. Closes #349.

## What proves it

Nine tests, each named for the bullet it answers. The acceptance criterion asks for a paint-path property and a transform mid-animation on the frame after first layout and arrived some frames later, for a pair differing only in whether the value carries an enter, and for nothing playing twice: `a_background_with_an_enter_starts_somewhere_else_and_arrives`, `a_scale_with_an_enter_starts_collapsed`, `an_enter_is_consumed_by_the_first_layout_and_does_not_play_again`.

The other six came out of the readings and each fails against the version that lacks its fix, verified by mutating the line rather than by argument: `an_enter_reaches_a_size_and_a_padding_too`, `a_text_colour_can_appear_from_somewhere_else`, `an_enter_that_shrinks_still_runs_the_forward_transition`, `an_enter_takes_what_the_property_takes`, `an_enter_on_a_timeline_refuses`.

The book chapter lost its enter section in #304 and has one back that is not transform-shaped; `mdbook test` compiles both samples.

## What the review found

**Three blocking findings, all fixed.**

An enter that *shrinks* ran the reverse transition and fired its `on_complete`. `animate_to` reads direction off the value it is leaving, and an enter has just placed that value at the start of the appearance. In the very pattern this feature exists for, that callback destroys the popup — so a menu appearing from a larger scale would have closed itself. An appearance now begins forward whichever way it travels.

Two of the four initialisers were proved by nothing: deleting the padding, border-width or text-colour enter left the whole suite green. Worse, the test named for padding asserted only on width. Both are now covered by assertions that observe the thing they name.

On an auto-height popup — the one named consumer — the enter was consumed during `measure_natural_size`, which runs before the surface is mapped and outside the frame window, so it also tripped the out-of-pass clock diagnostic at a guido internal. A measure pass is not an appearance and is now skipped.

Worth answering, both acted on: `entering_from` took `impl Into<T>`, which silently misses the three widenings guido declares with `converts!(S as T)` — it now takes `IntoVal<T>`, the crate's own conversion trait, so an enter is written in the same vocabulary every other position takes. And the unreachable panic arm is an `unreachable!` with the reasoning `into_eased` already uses, while the reachable one has a `#[should_panic]` test.

Notes acted on: `docs/STYLING.md` and the widgets skill both enumerated the motion vocabulary as two verbs, and the skill is what the next agent reads.

The review agreed one commit is right here: `Motion::Ease` becoming a struct variant breaks its constructors in the same file, so no split builds.

## What was checked by hand

Nothing. Everything is reachable from the unit tests, and no golden moved because no golden carries a first frame.

## The second commit

The mutation job turned red on this branch, and it was right to. Fifty-one mutants, twenty-one caught, twenty-nine unviable, **zero survivors** — and one timeout, which is what failed the job: `TextAnims::retarget` mutated to return a font size of `-1.0` did not finish in thirty seconds.

That is not an artefact. `text("hello").font_size(-1.0)` hangs guido on `main` too, reproduced by hand with the build completed first so the wait is the run. `Metrics` built from a non-positive size does not terminate inside cosmic-text, and layout reaches it before anything can draw — no panic, no frame, nothing said. One line of ordinary application code, and a bar that hangs takes the user's panel with it.

Four places build `Metrics` from a font size, so the clamp is one door they all pass through. `a_size_that_cannot_be_shaped_still_answers` runs on a thread with a deadline, because what it guards is a hang: asserted inline, a regression would take CI down instead of failing it. Verified by removing the clamp — it fails in ten seconds rather than spinning.

This is the first time that job has found something here rather than confirming what the tests already said.

## Left undone

**#347** — padding and border width have never reported drift. Their gate is keyed on a flag only `set_immediate` sets, and nothing had ever called it for those two, so the pull half of the resync invariant has been off for them. This change is what made it visible, by adding the first seed-time call that touches them; it is present without this change and needs an acceptance criterion this change's tests cannot state.

Not opened as issues, because nobody is worse off today: the state change named in the acceptance is not separately exercised, since `take()` makes the once-only structural rather than conditional. And the ashell port needs two edits rather than one to build against current guido — its popup calls `.transform(..)` as well, which #304 also removed, and no `transform` setter exists to replace it.

https://claude.ai/code/session_01Xb5BYhQqkrE8ngiyKE6FRX

### DIFF
--- a/.claude/skills/widgets/SKILL.md
+++ b/.claude/skills/widgets/SKILL.md
@@ -205,6 +205,21 @@ A bare number is milliseconds, eased out — the one place the animation
 vocabulary has two defaults, because `Transition::default()` is a spring and
 has no duration for a number to attach to.
 
+`.entering_from(..)` is the third verb, and a modifier on the first rather than
+a peer: a transition says how a value moves, an enter says where it starts the
+one time the widget appears, and CSS keeps the same two apart as `transition`
+and `@starting-style`. It plays once, at the first layout, and is consumed
+there.
+
+```rust
+container().background(theme.surface.transition(200.0).entering_from(Color::TRANSPARENT))
+```
+
+The take lives on `AnimationState`, not in the initialiser that seeds most
+properties — there are four of those, and an enter asked for in one of them
+compiles on every setter and works on half. That is #303's own defect, and the
+rule is that a property is *placed* in exactly one place, whichever it is.
+
 Three rules the shape depends on, in order of how easily they are broken:
 
 - **A declaration is the whole property.** Restating one replaces the value
--- a/book/src/transforms/animated.md
+++ b/book/src/transforms/animated.md
@@ -20,6 +20,50 @@ container()
 
 When `rotation_signal` changes, the transform animates smoothly.
 
+## Appearing mid-animation
+
+`transition` says how a value moves. `entering_from` says where it starts the
+one time the widget appears — the CSS pair, where `transition` supplies the
+motion and `@starting-style` supplies the before-value.
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+# let open = create_signal(true);
+# let collapsed = Scale::new(1.0, 0.0);
+container()
+    .scale(
+        (move || if open.get() { Scale::NONE } else { collapsed })
+            .transition(Transition::spring(SpringConfig::SNAPPY))
+            .entering_from(collapsed),
+    )
+    .pivot(Pivot::TOP)
+# ;
+# }
+```
+
+A menu that scales open on its first layout, and closes by the same transition
+running the other way.
+
+It is not transform-shaped: the verb hangs off the value, so it reaches every
+animatable property. A card can fade in.
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+container()
+    .background(Color::rgb(0.1, 0.1, 0.15).transition(200.0).entering_from(Color::TRANSPARENT))
+# ;
+# }
+```
+
+The enter plays once, at the first layout, and is consumed there: a relayout, a
+resize or a state change is not an appearance and does not replay it. It needs a
+transition to travel with, so declaring one on a timeline is a panic rather than
+a value quietly ignored — a timeline already says where it starts.
+
 ## Duration-Based Animation
 
 Standard easing curve transitions:
@@ -199,6 +243,9 @@ impl Container {
     pub fn rotate<M>(self, degrees: impl IntoAnimated<f32, M>) -> Self;
     pub fn scale<M>(self, factor: impl IntoAnimated<Scale, M>) -> Self;
 }
+
+// Where a property starts the one time its widget appears.
+Animated::entering_from(self, from: T) -> Animated<T>
 
 // Duration-based
 Transition::new(duration_ms: f32, timing: TimingFunction) -> Transition
--- a/docs/STYLING.md
+++ b/docs/STYLING.md
@@ -246,7 +246,8 @@ carries none — so it asks on its sources alone, and empties them to switch off
 
 **What is not reactive** is structural: `.layout(..)`, the axis a `.scroll(..)`
 is built with, `.control()`, and the motion a value is declared with —
-`.transition(..)` and `.timeline(..)`. These say what
+`.transition(..)`, `.timeline(..)` and the `.entering_from(..)` that says where
+a transition begins on the frame the widget appears. These say what
 kind of thing the container *is*; change one and you are describing a different
 widget, so declare it in the closure that builds the widget instead. The
 *value* a motion decorates is as reactive as any other.
--- a/src/animation/animated.rs
+++ b/src/animation/animated.rs
@@ -24,7 +24,7 @@
 //! .when_hovered(|s| s.background(HOT.transition(900.0)))   // does not compile
 //! ```
 
-use crate::reactive::{IntoSignal, Signal};
+use crate::reactive::{IntoSignal, IntoVal, Signal};
 
 use super::{Animatable, Keyframes, TransitionConfig};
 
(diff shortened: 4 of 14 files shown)


MalpenZibo:
## Padding and border width folded in, and #347 corrected

They were a fourth initialiser case: seeded at builder time, read at the first layout only for their subscription. That left them the two properties `is_initial()` was never false for, and the drift check in `resync_animation_targets` is gated on exactly that flag — so it could never speak for them.

They are now placed by the same call as the other nine. One helper instead of two, and it removes a split this change would otherwise have introduced: a padding *with* an enter would have been placed and a padding without one not, so `is_initial()` would have depended on whether an enter was declared. All 996 tests pass with the special case gone.

**I have to correct #347 as I wrote it.** The gate really is shut for those two, but the consequence I claimed — a write being lost — is not one I can produce. `advance_anim!` is handed a freshly recomputed target on every frame it runs, so the value converges without the drift check; what the check adds is the *wake*, for a settled container with no other reason to run a frame. I wrote a test for the untracked-branch case and it passed with the seeding removed, because the harness pumps frames unconditionally — so it proved nothing and I deleted it rather than keep a test whose name claims coverage it does not have.

The full correction is on #347. What remains there is narrower than its title and worth the maintainer's judgement rather than my guess.

**Left undone** in the body above should therefore read: #347 stays open, smaller than described, and no longer as a consequence of this change.

MalpenZibo:
## Why the mutation job is red, and why it is being merged that way

Six missed mutants, all inside `TextRenderState::prepare_layer` and `TextMaskRenderer::mask`. Both need a wgpu device. The mutation job builds the library and its tests but runs on a runner with no adapter, so the golden tests that would kill them skip — no test that job can execute will ever reach those lines.

They are not new code and they are not this change's code. They entered the mutated region because the clamp adds one line inside each of those functions, and `cargo mutants --in-diff` takes the whole enclosing function. Without the clamp there they would not appear; with it there they cannot be killed. That is the shape #278 already describes, and it is the friction `AGENTS.md` declines to impose when it says this job reports rather than blocks.

The clamp does have to be in all four places. With only the measurer clamped, layout completes and paint then hands the widget's own unclamped size straight to the draw path, which hangs there instead.

### What was removed before merging

An earlier revision added a `debug_assert!` to `shape` so that a mutant returning a bad size would panic rather than hang, turning a timeout into a caught mutant. It was removed. Its only real purpose was to satisfy the tool, and library code should not be written for that — the maintainer caught it and was right to.

What stays is the clamp, which fixes a hang reproduced by hand on `main`, and the line-height ratio folded into the same helper, which removes a constant that was written out four times. The timeout that remains is `shapeable_metrics` mutated to return a negative line height: it reintroduces the hang, so the suite times out rather than failing, which is the correct behaviour of a suite whose every text measurement has been broken.

### The record

| | before this change | now |
| --- | --- | --- |
| missed | 0 of 51 | 6 of 73, all wgpu-only |
| caught | 21 | 34 |
| timeouts | 1 — the hang, undiagnosed | 1 — the same hang, deliberately reintroduced by the mutant |

The first run's timeout is what found #349. That is the first time this job has told us something the tests did not already know.

Issue #351, opened 2026-09-10
A hidden container still animates, so an endless sequence on one renders every frame to nothing


MalpenZibo:
## Observed

`Container::advance_animations` has no visibility gate. `visible(false)` is checked in layout (`src/widgets/container/mod.rs:1455`) and in paint (`:1616`, `:1703`), both of which are downstream of the job that already woke the loop.

So this renders and presents the surface every frame, forever, with nothing on screen:

```rust
container()
    .visible(busy)
    .rotate(0.0.timeline(spin()))   // spin() repeats forever
```

`is_animating()` is `timeline.playing.is_some()` (`src/widgets/container/animations.rs:391`), which for `Repeat::Forever` is permanently true, so `advance_anim!` asks for an Animation job on every frame. The value changes nearly every frame, so the job carries `RequiredJob::Paint`, so the surface repaints and presents.

## Who is worse off

Anyone who reaches for `.visible(..)` rather than conditional mounting for something that animates endlessly — and the container API offers both, side by side, in `book/src/concepts/container.md`. The two now differ in a way nothing says: `child(move || busy.get().then(spinner))` settles, `spinner().visible(busy)` does not.

The cost is a bar rendering at vsync for as long as it runs, on battery, showing nothing. A culled child is the same shape: it is in the tree, so it is not "gone", but nothing it draws reaches the screen.

## Why it is separable from #348 and #213

`Repeat::Forever` (#213) is the first animation in guido that never settles, which is what makes this visible — but the gate is missing for every animation, not only that one. A transition on a hidden container has the same hole and merely stops asking sooner. The acceptance below is about visibility and applies to all of them, which is exactly what #213's own tests could not state.

The codebase already knows this shape. `src/jobs.rs:275-281` records the caret: *"a focused input pinned the loop at 60 fps and repainted 113 frames out of 114 to no effect. On a lock screen, where a field is focused by definition, that runs all night."* That was fixed with `request_job_at`. An endless sequence cannot be deadline-scheduled, but it can be gated on whether anyone would see it.

## Expected

A container that is not visible asks for no animation job.

## Acceptance

In `src/widgets/container/characterization.rs`, beside `a_value_that_is_not_a_number_lets_the_surface_go_idle`: a container with `visible(false)` carrying an endless sequence, driven for several frames, and then `assert!(!pump(&mut h))`. It fails today. The same container with `visible(true)` must still animate, which is what keeps the fix from being "never animate".

[...]

PR #352, opened 2026-09-10
A sequence plays because the widget is there, unless something asks it to wait

### LINKED ISSUE #213: Nothing can animate for an unknown length of time without the app owning frame timing
> As an application author, I want to show that something is happening for an
> amount of time I cannot predict, without owning frame timing myself.

### Confirmed in the code

A timeline is modelled as a *gesture*, and a wait is not one.

```rust
pub fn repeat(mut self, times: u32) -> Self { self.repeat = times.max(1); self }
pub fn total_ms(&self) -> f32 { self.duration_ms * self.repeat as f32 }
```

A repeat count is a number fixed when the timeline is built, and the whole thing
has a total duration. There is no "while this signal holds", so a large `repeat`
cannot be called off when the answer arrives — the spinner keeps turning after
the work is done.

And a widget that exists only for the length of the wait never plays at all:

```rust
pub(crate) fn set_play_trigger(&mut self, trigger: Signal<u32>) {
    if let Some(timeline) = &mut self.timeline {
        timeline.last_play = trigger.get_untracked();   // ← records the value now
        timeline.trigger = Some(trigger);               // ← and waits for a change
    }
}
```

`keyframes_transform(kf, plays)` starts on a *change* of `plays`, and the
starting point is whatever the trigger held when the widget was built. A spinner
mounted the instant `busy` flips is built with the trigger already at its new
value, so it sits perfectly still waiting for a change that has already happened.

The shape is right for a shake — a discrete gesture, played a fixed number of
times, triggered by an event that happens while the widget is already there. It
is the wrong shape for a wait, which starts when the widget appears and ends when
something else says so.

### What the application pays

allock steps the angle itself: a `tokio::select!` branch in the auth task, every
40 ms, gated on `answering || scanning`. It works, and it is the library's job
being done in application code — guido owns frame timing for every other moving
thing on that screen, and this one thing has to be driven from a task that also
talks to PAM.

### Done when

Either of two answers, and the second is probably the shorter one:

- **A timeline that can repeat while a signal holds, and be stopped mid-run.**
  `repeat_while(signal)` beside `repeat(n)`, and a play trigger that fires on
  mount when its condition is already true rather than only on a change. That
  second half is worth fixing regardless — "starts on a change" is a trap for
  anything conditionally mounted, not only for spinners.
- **A `spinner()` widget.** It settles the same need and also settles the step
  size, the rate, and the fact that a glyph only turns in place if its ink is
  centred on its line box — three decisions every application would otherwise
  make separately and differently.

A spinner also runs straight into #212: a rotation animation cannot be expressed
as a transform lerp, which is why allock's step is 12° and not something chosen
for how it looks. Whichever way this is answered, that one comes first.

MalpenZibo:
## What changed

An application could not show that something was happening for a length of time it could not predict without owning frame timing itself — allock still steps its spinner from the task that talks to PAM. A timeline was modelled as a gesture: `repeat(n)` is a count fixed at build time, and a sequence started on a *change* of its trigger never plays for a widget mounted at the instant its condition became true. So the default flips. A sequence plays because its widget exists; `played_by(trigger)` is how a gesture says it would rather wait; `timeline(keyframes)` takes one argument; and `Repeat::{Times, Forever}` replaces the bare count. Closes #213.

```rust
fn spin() -> Keyframes<f32> {
    Keyframes::new(1000.0).at(1.0, 360.0).repeat(Repeat::Forever)
}
container().child(move || busy.get().then(spinner))
```

## What proves it

Six tests in `src/widgets/container/characterization.rs`, each verified to fail against the code without its fix:

- `showing_the_widget_starts_the_sequence_and_hiding_it_stops_everything` — the whole design in one test, and the case the issue says fails today: the spinner is built *after* `busy` flips, then the surface goes idle when it goes away. This is what makes "no stop call" a design rather than an omission.
- `a_sequence_with_no_trigger_plays_because_it_exists` and `a_sequence_played_by_a_trigger_waits_for_it` — the two defaults, in both directions.
- `a_sequence_that_repeats_forever_is_still_running_much_later`, whose `Times(1)` control is what makes the claim about the repeat rather than about the clock.
- `seeding_a_sequence_never_asks_what_time_it_is` — the seed pass asks for the frame, the animate pass plays. Moving the play into the seed makes it fail.
- `an_endless_sequence_still_moves_frame_to_frame_after_a_long_time` — a month in, two consecutive frames still differ.

No golden or snapshot moved and none should: this changes *when* a sequence starts, not what a pixel looks like.

## What the review found

**One blocking finding, fixed.** The mount and unmount lifecycle the whole design rests on was proved in neither direction — no test removed the widget, and none mounted one after its condition changed. Both are now the first test above. Writing it also corrected my own understanding: a dynamically mounted child needs a reconcile pass first, so the sequence starts two frames after the flip rather than one.

Worth answering, both acted on: an endless sequence's elapsed time grew without bound, losing a whole 60Hz frame of resolution after about thirty-seven hours, and the ask-here/play-there split had nothing holding it apart.

Before that, four cleanup readings found that an earlier revision started the play inside `seed_animations`, which sat outside the measure-pass guard its neighbour has and read the trigger inside the *parent's* layout tracking scope. Removing that second play site deleted both at once. They also caught `played_by` narrowing from five accepted spellings to two, which is the rule this repository's own skill states.

## What was checked by hand

Nothing. `examples/keyframes_example.rs` still demonstrates only the triggered play, so there is nothing new to look at by hand without adding an example, and the behaviour is fully reachable from the tests.

## Left undone

**#351** — a hidden container still animates, so `spinner().visible(busy)` renders every frame to nothing while `child(move || busy.get().then(spinner))` settles. The gate is missing for every animation, not only endless ones; `Repeat::Forever` is simply the first that never stops asking. Its acceptance is about visibility and applies to all of them, which is why this change's tests cannot state it.

`Repeat::Times(0)` is writable and means once, the clamp having moved from `repeat()` to `total_ms()`. Documented as "at least once"; no issue, since nobody is worse off.

https://claude.ai/code/session_01Xb5BYhQqkrE8ngiyKE6FRX

### DIFF
--- a/.claude/skills/widgets/SKILL.md
+++ b/.claude/skills/widgets/SKILL.md
@@ -198,7 +198,7 @@ available on everything a property setter already accepts, and each returns an
 container()
     .background(theme.surface.transition(200.0))
     .width((move || if open.get() { 520.0 } else { 120.0 }).transition(SpringConfig::SNAPPY))
-    .rotate(0.0.timeline(shake(), rejections))
+    .rotate(0.0.timeline(shake().played_by(rejections)))
 ```
 
 A bare number is milliseconds, eased out — the one place the animation
--- a/book/src/animations/keyframes.md
+++ b/book/src/animations/keyframes.md
@@ -21,8 +21,8 @@ container()
             .at(0.15, 2.0)
             .at(0.40, -1.6)
             .at(0.65, 0.9)
-            .at(1.0, 0.0),
-        rejections,
+            .at(1.0, 0.0)
+            .played_by(rejections),
     ))
 # ;
 # }
@@ -42,8 +42,8 @@ container()
         Keyframes::new(240.0)
             .at(0.0, surface)
             .at(0.3, Color::rgb(0.8, 0.2, 0.2))
-            .at(1.0, surface),
-        errors,
+            .at(1.0, surface)
+            .played_by(errors),
     ))
 # ;
 # }
@@ -65,7 +65,7 @@ sequence resting on a live signal has nowhere else to put its expression:
 # let shake = || Keyframes::new(320.0).at(0.0, 0.0).at(0.5, 2.0).at(1.0, 0.0);
 // A glyph that spins on one signal, and shakes on another.
 container()
-    .rotate((move || spin.get() as f32 * 90.0).timeline(shake(), rejections))
+    .rotate((move || spin.get() as f32 * 90.0).timeline(shake().played_by(rejections)))
 # ;
 # }
 ```
@@ -75,16 +75,51 @@ persist is a transition.
 
 ## What plays it
 
-The second argument is a signal, and **every change to it plays the sequence
-once**. A count rather than a flag, because two refusals in a row are two
-events and a signal that stays equal notifies nobody — the second wrong
+**A sequence plays as soon as the widget carrying it exists.** That is what a
+wait looks like: a spinner should turn from the moment it is on screen, and
+there is nothing to ask it.
+
+**`played_by(trigger)` makes it wait instead**, and then every change to that
+signal plays it once. A count rather than a flag, because two refusals in a row
+are two events and a signal that stays equal notifies nobody — the second wrong
 password has to shake as loudly as the first. SwiftUI's keyframe animator takes
 its trigger the same way.
 
-Nothing plays on the first frame: the container remembers what the signal held
-when it was built. Played again while it is still running, a sequence starts
-over from the top rather than continuing — half a shake is not what a second
-refusal means.
+A sequence with a trigger plays nothing on the first frame: the container
+remembers what the signal held when it was built. Played again while it is
+still running, a sequence starts over from the top rather than continuing —
+half a shake is not what a second refusal means.
+
+## How long it runs
+
+`repeat(Repeat::Times(n))` plays the whole run `n` times and then rests.
+`Repeat::Forever` never stops:
+
+```rust
+# extern crate guido;
+# use guido::prelude::*;
+# fn main() {
+# let busy = create_signal(true);
+fn spin() -> Keyframes<f32> {
+    Keyframes::new(1000.0).at(1.0, 360.0).repeat(Repeat::Forever)
+}
+
+fn spinner() -> Container {
+    container().rotate(0.0.timeline(spin())).child(text("*"))
+}
+
+// On screen only while the work is outstanding.
+container().child(move || busy.get().then(spinner))
+# ;
+# }
+```
+
+There is no way to stop an endless sequence other than removing the widget that
+carries it, and that is the point. The widget being on screen *is* the
+animation running, so showing and hiding it is the only switch to keep in step
+— nothing to call off, and no way to leave a spinner turning after the work is
+done. Compose reaches the same place from the other direction: an infinite
+transition there runs for as long as the composable is in composition.
 
 ## Offsets and easing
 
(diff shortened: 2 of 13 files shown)


MalpenZibo:
## Second commit: the visibility gate, and a correction to #351

Folded in on request. #351 is closed by this pull request rather than left open.

**I had to correct the issue first, because I wrote it without testing it.** It claimed two cases and only one exists.

A container hidden *before* its animation starts was never affected: `Container::layout` returns at its own visibility gate before reaching `seed_animations`, so the sequence is never seeded and nothing is ever asked for. The test I wrote for that case passed against unfixed code, which is how I found out.

A container hidden *while* its animation runs was the real one. `advance_animations` had no gate, and `visible(false)` is checked only in layout and paint — both downstream of the job that already woke the loop.

**Why it belonged here after all.** Every animation before `Repeat::Forever` settled by itself, and a settled animation stops asking. So this was a bounded waste until the first commit made it unbounded. Splitting them would have meant shipping the unbounded version first.

**The fix is five lines.** `advance_animations` reads `visible` under tracking, exactly as `layout` reads the same signal, and returns early. The tracked read *is* the resumption: flipping back to visible queues the Animation job the pass would otherwise have had to keep asking for. `a_hidden_container_with_an_endless_sequence_lets_the_surface_go_idle` asserts both directions, because a gate that never reopens is just a stop.

**Left open, and not by this change:** a child culled out of a scroller is invisible to the user, present in the tree, and `visible` is true — so it still advances. That is a different notion of "not seen" and wants its own answer. Recorded on #351 rather than reopened, since its title is about visibility.

Both commits verified green on their own.

MalpenZibo:
## The mutation job, and the one that survives

It found something, and it took three rounds to answer it honestly.

| run | missed | caught | unviable |
| --- | --- | --- | --- |
| first | 8 | 39 | 25 |
| second | 2 | 44 | 25 |
| third | **1** | 37 | 23 |

**The first run was right about my tests.** Eight missed, all in plain logic added by this change — not the unreachable GPU code of the previous pull request. Three of them exposed the same weakness: `an_endless_sequence...` asserted only that the animation *moved*, and every arithmetic mutation in the reduction kept it moving.

Writing the stronger assertion — that a whole number of runs later is the *same point in the run* — found a real defect. Reducing the elapsed time in `f32` was not enough on its own: at thirteen million runs the arithmetic is coarser than a whole run, so a month in, the sequence sat at the wrong point in its cycle rather than merely being imprecise. It reduces in `f64` now, and the stored rebase is gone: the reduction is recomputed from the start every frame anyway, so keeping one was a second place for the same arithmetic to be wrong in. That deleted a mutant rather than testing one.

Three tests were added and each was verified against the mutant it is for: the phase assertion, `a_finished_sequence_stops_asking_to_be_played`, and `a_property_with_no_sequence_asks_for_no_play`.

### The survivor

`duration > 0.0` → `duration >= 0.0`, in the guard that keeps an endless sequence from dividing by its own length when that length is zero. What such a sequence displays is identical either way — the sample falls through to a finite value on both paths — so the mutant is equivalent by construction rather than untested. The comment at the site says so.

### What I removed

An earlier revision of this comment would have said a zero-length run was fixed too. It is not. I added a guard for it and justified it three times — as preventing a NaN, as letting the surface settle, as changing the drawn value — and each justification was wrong: the test behaved identically with and without the guard every time. It is removed. A change I cannot write a distinguishing test for does not ship as though it were proven, and the survivor above is the honest price of that.

